PR #11, opened 2016-12-09
Fix crash when debug is enabled

gburca:
- If the error or info log files can not be created, use stderr and
  stdout instead.


gburca:
I don't remember exactly how the crash happened, but when linking an app (MediaTomb) with the debug version of the library and running it from a directory in which the user didn't have write permissions, the app would crash. I traced it down to the libupnp logging code which was trying to write to a NULL log file handle in `UpnpPrintf()`.

mrjimenez:
Seems ok, will be committed, thanks!

PR #19, opened 2017-02-25
Remove NO_DEBUG define

jcowgill:
The define is never used anywhere.


ffontaine:
Good catch, it also seems that NDEBUG is never used anywhere.
We could also replace UPNP_HAVE_DEBUG and DEBUG defines by a single define as both are changed at the same time.

PR #20, opened 2017-02-25
Replace MD5 impmplementation with a public-domain version

jcowgill:
Currently the RSA MD5 implementation is used. Unfortunately the license has some potential issues:
* The license does not explicitly allow distributing derivative works. This was the original argument used in [Debian #459516](https://bugs.debian.org/459516).
* The license contains an advertising clause similar to the BSD 4-clause license. This is incompatible with the GPL and if it were enforced, would require RSA to be mentioned by pretty much everyone who uses pupnp.

The simple solution is to replace it with a public domain implementation. I've taken OpenBSDs implementation and tweaked it slightly for use by pupnp.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,6 +2,28 @@
 Version 1.8.1
 *******************************************************************************
 
+2017-07-07 James Cowgill <james410(at)cowgill.org.uk>
+
+	Replace MD5 impmplementation with public-domain version
+
+	Currently the RSA MD5 implementation is used. Unfortunately the license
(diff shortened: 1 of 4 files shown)


mrjimenez:
Hi James,

Would you mind including an entry in Changelog (just copy the commit message) and maybe detail in which way you tweaked the original implementation?

Regards,
Marcelo.

jcowgill:
Updated - does that look good?

mrjimenez:
Perfect!

PR #23, opened 2017-03-07
Add option to use REUSEADDR on miniserver socket

v00d00:
This adds a compile time option "--enable-reuseaddr" which enables the
previously unavailable SO_REUSEADDR codepath in miniserver.c

This allows us the normal advantages of REUSEADDR when implementing
servers that would like graceful restarts.

See v00d00/mediatomb#34

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -496,6 +496,11 @@ if test "x$enable_postwrite" = xyes ; then
         AC_DEFINE(UPNP_ENABLE_POST_WRITE, 1, [see upnpconfig.h])
 fi
 
+RT_BOOL_ARG_ENABLE([reuseaddr], [no], [bind the miniserver socket with reuseaddr to allow clean restarts])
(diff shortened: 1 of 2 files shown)


mrjimenez:
Hi Ian,

I don't know if I am missing something on the interface of github, but the patch seems incomplete to me. I see that you create a variable called reuseaddr_on, but I don't see the use in the patch. On the other hand, the file in your repository seems to have the code that uses it.

Do you have to do other pull requests? Shouldn't them be merged? Or is it my mistake?

Regards,
Marcelo.

v00d00:
Hi Marcelo,

That is indeed the odd thing _the code is already there_, there was just no way to use it!

You can see previously it was only ever set to `0`.

This simply adds a flag to enable a code-path that has been there a long time...

mrjimenez:
Amazing... :)

Thanks for the patch!

PR #27, opened 2017-03-08
Fix for SO_REUSEADDR code path

whorfin:
Thanks much for #23, but there was a problem.  Unsurprising as it was ancient code.
This is fixed and tested.  I made the equivalent change to the IPV6 code path, but haven't tested.
This fixes https://github.com/v00d00/mediatomb/issues/34

There's also a commit here to avoid configure hassling over autogenerated files, feel free to drop that part if it offend thee

Cheers


ffontaine:
I'm ok to remove autogenerated files, I think these files were added for the Windows build system (which does not use autotools) but with latest update on threadutil, I think it is now broken until someone fix it.

v00d00:
Only thing I notice is that the white-space is tabs not spaces in that file?

mrjimenez:
Hi Fabrice, if I recall correctly, you are right. Since the windows build is broken anyway, I think we can do away with these files until someone fixes it.

PR #31, opened 2017-04-19
Cleaning of project files 

jbkempf:
Those simple commits are removing files that are not used anymore, notably old VC files, TODO and svnignore.


mrjimenez:
Hi Jean-Baptiste,

Before accepting the changeset, let me ask a question to any windows developer that might be reading this. Is it ok to remove support for vs 2005, vs 2008, vc 8 and vc 9?

Also, are the vc 10 files ok for a current windows build?

Unless someone responds with a good reason not to do so, I will accept the patch.

Regards,
Marcelo.

jbkempf:
IMHO, even vc10 studio files are outdated. But they should be openable with a recent VS.

VS2015 and VS2017 are way more interesting, because they have a community edition that is free (as in beer).

PR #33, opened 2017-04-19
upnpapi: Fix default interface index

chouquette:


mrjimenez:
Hi Hugo,

Can you explain why this is necessary? Is this causing a bug or is it just an aesthetic change?

Regards!

chouquette:
Hi,

This was aiming at fixing discovery on windows, but as I'm re-reading https://msdn.microsoft.com/en-us/library/windows/hardware/ff570824(v=vs.85).aspx I realize I mistaken `A value of zero specifies the default zone.` to be related to the `sin6_scope_id` field, while it seems to be related to `SCOPE_ID::Zone`

Moreover, the testing showed it was not enough to fix things, so please disregard this patch.

Going forward, I believe the `sin6_scope_id` field holds a very different meaning from the one in GNU/Linux, and will need some fixing, but this patch ain't the correct one.

mrjimenez:
Closing at your request.
Thanks for your attention.

PR #38, opened 2017-05-20
Fix build error caused by backslash blank line

SandPox:


SandPox:
oops, this already get a PR by #36

PR #40, opened 2017-05-26
Don't append flags like -O0/-Os which clobber the user's settings

v00d00:
This is patch we carry in Gentoo, its generally frowned upon to hard code these flags.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -567,25 +567,6 @@ AC_PROG_EGREP
 echo "--------------------- Default compilation flags -------------------------------"
 echo host is $host
 echo host_os is $host_os
-case $host_os in
-freebsd*)
-	echo "Using FreeBSD specific compiler settings"
-	# Put FreeBSD specific compiler flags here
-	;;
-*)
-	echo "Using non-specific system compiler settings"
-	if test x"$enable_debug" = xyes; then
-		# AC_PROG_CC already sets CFLAGS to "-g -O2" by default,
-		# but only if CFLAGS was not previously set.
-		#:
-		# Use -O0 in debug so that variables do not get optimized out
-		AX_CFLAGS_GCC_OPTION([-O0, -g])
-	else
-		# add optimise for size
-		AX_CFLAGS_GCC_OPTION([-Os])
-	fi
-	;;
-esac
 
 AX_CFLAGS_WARN_ALL
 echo "-------------------------------------------------------------------------------"


mrjimenez:
Is there anyone opposed to this? I have only one question, this apparently breaks ./configure --enable-debug, so what is the suggestion in this case?

ffontaine:
I'm not opposed to remove "-O0" and "-Os" however removing "-g" will indeed break one of the effect of --enable-debug. Indeed, this option has currently three effects:
- enable calls to assert funcion
- enable log files
- enable "-O0 -g" so that the user can use gdb or any other debugger

This is not totally uncommon to append -g in debug mode but we can remove it if we document it in the ChangeLog.

mrjimenez:
What is the sane "gentoo way" to configure these flags? I am not happy with breaking --enable-debug for no reason. If there is a good reason, there must be a sane alternative.

As for "-Os" no longer being default, the resulting effect is that builds of the library will start to get bigger on the new releases. Since libupnp is used in embedded systems, this is not a nice side effect too.

I would rather keep things the way they are, and add a --default-optimization/--no-default-optimization in configure, "--default-optimization" beeing the default "-Os" as it is today, gentoo builds would run configure with "--no-default-optimization", which apparently will not conflict with "user settings" as stated in the patch, what do you think?

v00d00:
I think the embedded case would already have set up their toolchain/flags as they want already, as its important to them, they probably have all kinds of stuff set up in the pipeline for different libc and things too.

The nice way is not to set these flags at all, that way, embdedded people can set the flags they want, Gentoo people can set the flags they want also.

as @ffontaine mentions, most libraries specifies to build with -g if you are debugging, its pretty expected.

(In gentoo these flags are set in [make.conf](https://wiki.gentoo.org/wiki//etc/portage/make.conf))

ffontaine:
I think that the proposal of @mrjimenez is also nice as it won't break anything and should fit the needs of gentoo or any other build system.

v00d00:
I don't mind the idea, but I think it could be more simple: if the CFLAGS is set in the environment then respect those, otherwise, just use the logic that was present before?

ffontaine:
This is not so simple, most of the open source projects that I know set their own CFLAGS. Some of them only enable a few options such as "-g" in debug mode or "-std=c99" but others make heavy configuration such as cjson which enables the following by default: 
```
option(ENABLE_CUSTOM_COMPILER_FLAGS "Enables custom compiler flags for Clang and GCC" ON)
if (ENABLE_CUSTOM_COMPILER_FLAGS)
    list(APPEND custom_compiler_flags
        -std=c89
        -pedantic
        -Wall
        -Wextra
        -Werror
        -Wstrict-prototypes
        -Wwrite-strings
        -Wshadow
        -Winit-self
        -Wcast-align
        -Wformat=2
        -Wmissing-prototypes
        -Wstrict-overflow=2
        -Wcast-qual
        -Wundef
        -Wswitch-default
        -Wconversion
        -Wc++-compat
        -fstack-protector-strong
        -Wcomma
        -Wdouble-promotion
        -Wparentheses
        )
endif()
```

v00d00:
Some flags would need to be set always eg `-stdc89` are fine, its just the optimisation flags we care about in particular.

v00d00:
Lets close this for now.

PR #44, opened 2017-06-26
Device eventing: add possibility to discard events based on age

medoc92:
Add possibility to discard events destined to non-responding control point based on event age, in addition to maximum queue lenth.

- Delete eventKey field from event data structure: it was set but never used.
- Add ctime age field.
- Add max age configuration parameter.
- Define function for setting max queue size and max event age, in the case where compiled-in defaults are not appropriate.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -512,11 +512,6 @@ if test "x$enable_unspecified_server" = xyes ; then
         AC_DEFINE(UPNP_ENABLE_UNSPECIFIED_SERVER, 1, [see upnpconfig.h])
 fi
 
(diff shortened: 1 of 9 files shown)


v00d00:
I wish my projects got patches like this! 👍

medoc92:
Thanks, but I hope you can give it a bit of testing. A simple way to reproduce event non-processing is to suspend a control point.

Also I did not mean the second change about accept-ranges to be part of the same pull-request. I should have created another branch. I guess it's simple enough to be sorted out. This was originated from Damien Plisson (Audirvana), who was trying to stream to MPD using the internal server and saw that seeks were not working.

PR #46, opened 2017-07-16
extra headers

philippe44:
Allow extra headers to be sent to client in the File_Info struct by adding extra_headers, an array of struct Extra_Headers.

It contains "name" and "value" and the callee can populate a DOMString "resp" to add a header as well in the reponse

### DIFF
--- a/upnp/inc/upnp.h
+++ b/upnp/inc/upnp.h
@@ -799,6 +799,16 @@ struct Upnp_Subscription_Request
 
 };
 
+struct Extra_Headers
+{
+	/** The length of the file. A length less than 0 indicates the size
+	*  is unknown, and data will be sent until 0 bytes are returned from
+	*  a read call. */
+	char *name;
+	char *value;
+	DOMString resp;
+};
+
 struct File_Info
 {
 	/** The length of the file. A length less than 0 indicates the size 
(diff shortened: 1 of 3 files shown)


mrjimenez:
I see no problem with this patch, there is no API breakage, just API extension, since the File_Info struct is extended. If there is no objection, I will commit it in the next days.

Also, Philippe, are you interested in doing a similar patch for 1.8.x? The code should be almost the same, but the template objects do not currently support arrays within the class. It should be rather straightforward to add it.

Regards,
Marcelo.

philippe44:
Yes, I can do that as well. Do you have some non-regression tests to run? I want to make sure I've not done anything silly in importing that from 1.6.19. As well, in term of coding criterias, some thing are a bit basics like the number of extra headers is limited to 128. 

It could be dynamic & more elegant, but when I did that, it was answering the need while remaining quick & simple with a very small memory impact. Maybe the number of headers could be a visible define when rebuilding the library. 

But if you are confortable with this solution, then that's perfect

mrjimenez:
I tried to do a unittest of the template structure, take a look at upnp/unittest/templates.

I believe that if you use linked lists, you will already have template support:

EXPAND_CLASS_MEMBER_LIST(CLASS, TheList)

Well, its your code, you choose it. :)

Regards,
Marcelo.

philippe44:
I'm not very familiar with the rules of conducts when contributing to other's projects but I'll keep it like that then :-) I might do better implementation later, but the interface would remain the same

mrjimenez:
Don't worry too much.

The catch is: if you use arrays, you'll need to create a new "type" in the template mechanism. If you use lists, you have the support.

Arrays are fine and clean, so if you prefer, please keep your implementation.

Regards,
Marcelo.

v00d00:
No love for 1.8? :(

philippe44:
I will do something in the coming days

PR #54, opened 2017-09-07
Add userdata/cookie to virtualDir callbacks

v00d00:
As with the main Device APIs (UpnpRegisterRootDevice etc), it is useful to
have a userdata/cookie pointer returned with each callback.

This patch allows one cookie per registered path which enables a variety of
functionality in client apps

### DIFF
--- a/upnp/src/api/upnpapi.c
+++ b/upnp/src/api/upnpapi.c
@@ -4130,7 +4130,7 @@ int UpnpSetWebServerRootDir(const char *rootDir)
 #endif /* INTERNAL_WEB_SERVER */
 
 
-int UpnpAddVirtualDir(const char *newDirName)
+int UpnpAddVirtualDir(const char *newDirName, const void *cookie, const void **oldcookie)
 {
     virtualDirList *pNewVirtualDir;
     virtualDirList *pLast;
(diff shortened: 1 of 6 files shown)


upnp/src/api/upnpapi.c
```diff
@@ -4162,6 +4162,9 @@ int UpnpAddVirtualDir(const char *newDirName)
     while( pCurVirtualDir != NULL ) {
 	/* already has this entry */
 	if( strcmp( pCurVirtualDir->dirName, dirName ) == 0 ) {
+            if (oldcookie != NULL)
+                *oldcookie = pCurVirtualDir->cookie;
+            pCurVirtualDir->cookie = cookie;
```
ukleinek: these lines are indented using space while the surrounding lines use a mixture of tab and space :-\

v00d00: The indentation changes all over the place. I did try to to match mostly

mrjimenez: I know the code is still tab inconsistent, historically I have tried to make it consistent as I go through the files I change.

Just for your information, I think it is ok to change the whole file indentation to something sane whenever a patch for that file is coming ahead. Just make sure the white space patch comes separate.

Regards!

PR #59, opened 2017-12-05
Fix segmentation fault in http_MakeMessage

joerg-krause:
When upmpdcli is linked with libupnp 1.6.24 it aborts with an segmentation fault, but linking with libupnp workes just fine.

git bisect shows that commit 9c2e8ec8a0291ebe81959009e2f78edbdb47ced5 is the bad one as the variable `extras` is not properly checked before attempting to used it as a pointer.

Asure `extras` is not Null before using it.

### DIFF
--- a/upnp/src/genlib/net/http/httpreadwrite.c
+++ b/upnp/src/genlib/net/http/httpreadwrite.c
@@ -1662,14 +1662,16 @@ int http_MakeMessage(membuffer *buf, int http_major_version,
 			struct Extra_Headers *extras;
 			/* array of extra headers */
 			extras = (struct Extra_Headers *) va_arg(argp, struct Extra_Headers *);
-			while (extras->name) {
-				if (extras->resp) {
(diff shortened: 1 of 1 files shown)


v00d00:
Why an assert over a regular null check?

joerg-krause:
@v00d00 You're right! I've updated the PR.

mrjimenez:
Hi joerg-krause,

Just committed, thanks!

Would you mind also updating the Changelog?

Best regards,
Marcelo.

joerg-krause:
@mrjimenez Thanks for merging!

The ChangeLog has not entry for the 1.6.x branch after 1.6.22. Should I just add an entry for 1.6.25?

mrjimenez:
Hi Jörg,

I did a mistake in the last 1.6 release, and in the process of fixing it I forgot to update the Changelog section. I just updated it, should be ok now, thank you for pointing me that.

Best regards,
Marcelo.

joerg-krause:
@mrjimenez Thanks! I'll update the ChangeLog now.

PR #70, opened 2018-03-06
Minor docstring fixes

gabrielivascu:


ukleinek:
Given that both changes touch the same comment I would have done both changes in a single commit. But other than that:
Acked-by: Uwe Kleine-König <uwe@kleine-koenig.org>

gabrielivascu:
> Given that both changes touch the same comment I would have done both changes in a single commit.

@ukleinek I can squash them if you like.

PR #98, opened 2018-12-23
Extra Headers Improvements

v00d00:
- Rename ExtraHeaders to UpnpExtraHeaders
- Make UpnpExtraHeaders public (includes, exports)
- Populate the ExtraHeaders in the finfo passed to the VirtualDir get_info callback

With these patches I was sucessfully able to read and write extra headers from [my application](https://github.com/gerbera/gerbera).

Please review :+1:


mrjimenez:
There is some pending "const" stuff, but overall awsome job!

mrjimenez:
Should this work go into 1.8 or do we leave it only in master? I think it should go into 1.8, since it is ExtraHeaders stuff. The only issue (non-issue?) is that it adds stuff to the public API.

Opinions are welcome.

Regards,
Marcelo.

v00d00:
Thanks for merging.

It should go into 1.8 for sure, as with the extra headers stuff.

VVorttexx:
Well guys, are we finally going to be able to watch something on our Samsung TV? 
The NOT adding HTTP header is really frustrating.

PR #100, opened 2019-01-13
Add Request level cookies to virtualdir calls

v00d00:
Currently there is no mechanism inside libupnp to track data accross the `get_info` -> `read` ->` close` flow of VirtualDir calls.

This means that client applications may have to do operations such as parsing and access checks etc twice, once in get_info, and again in read, or somehow track request lifetimes.

This change adds a new argument to the VirtualDir callbacks, so is not backwards compatible.

### DIFF
--- a/upnp/inc/upnp.h
+++ b/upnp/inc/upnp.h
@@ -2571,12 +2571,14 @@ typedef void *UpnpWebFileHandle;
  * \brief Get-info callback function prototype.
  */
 typedef int (*VDCallback_GetInfo)(
-		/*! [in] The name of the file to query. */
-		const char *filename,
-		/*! [out] Pointer to a structure to store the information on the file. */
-		UpnpFileInfo *info,
-		/*! [in] The cookie associated with this VirtualDir */
-		const void *cookie);
+    /*! [in] The name of the file to query. */
+    const char *filename,
+    /*! [out] Pointer to a structure to store the information on the file. */
+    UpnpFileInfo *info,
+    /*! [in] The cookie associated with this VirtualDir */
+    const void *cookie,
+    /*! [out] The cookie associated with this request */
+    const void **request_cookie);
 
 /*!
  * \brief Sets the get_info callback function to be used to access a virtual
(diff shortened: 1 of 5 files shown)


mrjimenez:
Nice.

If I am not mistaken, this API is quite subject to changes, as for the following comment in upnp/src/inc/webserver.h:

`/* Later few more member could be added depending
￼ 	 * on the requirement.*/		￼ 	 * on the requirement.*/`

Wouldn't it be nice if we start using a handle parameter, just like everywhere else in libupnp, so that we did not change the function signature? That way, we would have less breakages, only API aditions.

Just a suggestion for a future patch.

Regards,
Marcelo.

v00d00:
That sounds like a great idea. I wouldn't have the foggiest idea where to start, but it sounds good.

medoc92:
Two remarks about this commit:

 - There are many whitespace changes. Please use the same indentation
   standard as the rest of the code !
   * Whitespace changes make the diffs difficult to read on github
   * If we end up with a code base with mixed indentation
     (tabs/spaces/amounts), everything will be more difficult in the
     future. The current style in libupnp is not my preferred one, but I
     adapt because it is more important to keep things consistent.
   I think that this is sufficiently important that the whitespace should have
   been corrected before the commit.

 - This will force one more set of ifdefs in client code ! So I'm really
   all in favor of stopping changing function signatures !

On a more general note, I wonder how wise it is to add more and more
functions to the internal web server. Users who need a full-fledged web
server can use a dedicated general purpose web server package
(e.g. microhttpd). This is what I do.

A full-fledged web server is a big thing to maintain (esp. given the shaky
nature of the current code base).

If we end up with an internal Apache, we will also end up with the
maintenance load and security issues which Apache generates, and I don't
think that we have the resources for this.


Marcelo Roberto Jimenez writes: > Nice.
 > 
 > If I am not mistaken, this API is quite subject to changes, as for the
 > following comment in upnp/src/inc/webserver.h:
 > 
 > /* Later few more member could be added depending ￼ * on the requirement.*/ ￼
 > * on the requirement.*/
 > 
 > Wouldn't it be nice if we start using a handle parameter, just like everywhere
 > else in libupnp, so that we did not change the function signature? That way,
 > we would have less breakages, only API aditions.
 > 
 > Just a suggestion for a future patch.
 > 
 > Regards,
 > Marcelo.

v00d00:
@medoc92 I actually purposefully made the whitespace changes. It is now consistently 4 spaces instead of a spaces/tabs mixture before.

I agree about the internal webserver and I would be hesitant to add much more.

medoc92:
(removed some trailing quote I had in my initial comment).

@v00d00 : upnp.h is still full of tabs :)
4 spaces is also my preferred style, but, in its vast majority, libupnp is indented with tabs, and I think that it would be simpler to stick with them.

PR #108, opened 2019-03-08
libupnp.pc.in: move openssl in Libs.private

ffontaine:
openssl is a private library so move it from Libs to Libs.private
in libupnp.pc.in


mrjimenez:
Hi Fabrice,

Should we backport it to 1.8?

Regards,
Marcelo.

ffontaine:
Hi Marcelo,

Yes, it should be backported to 1.8 (as well as #105).

Best Regards,

Fabrice

PR #111, opened 2019-04-07
fix off-by-one overread on global variable

smestdagh:
when c is large enough, m can become equal to sz (found with AFL).


mrjimenez:
Good catch!

PR #120, opened 2019-10-06
upnp/sample: Use UpnpResolveURL2 in samples

willnewton:
UpnpResolveURL is a tricky API to use correctly and the samples
allocate one bytes less than strictly required. This code has
been copied into other applications that similarly have problems
alocating the correct amount of space. Fortunately UpnpResolveURL2
is available and it avoids the need to allocate space in the caller
and does less copying too.


mrjimenez:
Thanks!

Maybe we should break the API for 1.10.x and stop supporting UpnpResolveURL(), any oppinions?

Regards!
Marcelo.

willnewton:
Yes I think deprecating the API would be a good idea. The cost for users to
update seems pretty small.

PR #122, opened 2019-11-20
Fix format string for ExtraHeaders

KarlStraussberger:
Crashes Gerbera when creating response for file request


whyman:
Why doesnt seem to crash on 1.8? It appears to be the the same code?

KarlStraussberger:
Did not test it in 1.8.5. Possibly the same issue there also.
Only becomes active when ExtraHeaders are enabled.

whyman:
Confirmed that the crash also happens in 1.8.5 :-1: 

But also that this patch fixes it there also :+1: 

@mrjimenez Can we have new releases, please :grin:

mrjimenez:
> I've tested and confirmed that the crash also happens in 1.8.5 
> 
> But also that this patch fixes it there also 
> 
> @mrjimenez Can we have new releases, please

Sure!

PR #123, opened 2019-11-25
Check for -lpthread too

ePirat:
Fixes missing linkage to pthread library


mrjimenez:
What about pulling a new upstream version?
Home page: https://www.gnu.org/software/autoconf-archive/ax_pthread.html
Source code: http://git.savannah.gnu.org/gitweb/?p=autoconf-archive.git;a=blob_plain;f=m4/ax_pthread.m4

ePirat:
@mrjimenez Good idea, done that instead.

PR #125, opened 2019-11-26
Fix interface filtering for Windows

ePirat:
The mentioned fix is not only needed for MSVC, but when
cross-compiling for Windows with mingw-w64 too.

### DIFF
--- a/upnp/src/api/upnpapi.c
+++ b/upnp/src/api/upnpapi.c
@@ -3360,9 +3360,8 @@ int UpnpGetIfInfo(const char *IfName)
 		}
 		if (ifname_found == 0) {
 			/* We have found a valid interface name. Keep it. */
-#ifdef UPNP_USE_MSVCPP
 			/*
-			 * Partial fix for VC - friendly name is wchar string,
+			 * Partial fix for Windows: Friendly name is wchar string,
 			 * but currently gIF_NAME is char string. For now try
 			 * to convert it, which will work with many (but not
 			 * all) adapters. A full fix would require a lot of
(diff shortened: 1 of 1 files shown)


mrjimenez:
Doesn't that break UPNP_USE_BCBPP code?

ePirat:
Thats a good question, I don't know what `UPNP_USE_BCBPP` is as it seems not documented anywhere, so I am clueless how to test this, though.

Do you know what that define is and when it will be define?

mrjimenez:
I don't have windows here, but I guess that this should somehow be defined when using the Borland C compiler. If you add the fix for mingw on the ifdef and left the #else part intact, that should be ok.

ePirat:
Does this still work at all? Borland is ancient afaik and nothing in libupnp seems to ever define `UPNP_USE_BCBPP` at all.

mrjimenez:
IRC, this used to be defined externally.

I agree this is old code. The point I want to make is this: the code is "#if UPNP_USE_MSVCPP". The "#else" part means "all other compilers go here". I am no windows expert to give opinions on that, but apart from Borland outdated compilers, aren't there other C compilers for windows that should go in the else part? If that is the case, then your patch is ok, but as a "non-windows coder", but "other stuff coder", that "wcstombs" thing seems odd and until now has been kept as a MSVC only thing.

If you tell me it is not a particular compiler thing, but a Windows thing, I will gladly accept your patch, since it makes the code simpler. :)

Regards,
Marcelo.

ePirat:
`wcstombs` is part of the UCRT and the MSVCRT as far as I can see, so should work with any compiler for Windows using these windows runtimes. It is present in mingw-w64 too, so works when cross compiling with gcc or clang too.

mrjimenez:
Ok, good enough then. Thanks!

PR #127, opened 2019-11-26
Use rand_s in get_random_info on Windows

ePirat:


mrjimenez:
Why are you doing this? I can agree that the code looks simpler, but is there another reason to change this?

ePirat:
It provides "proper" randomness (cryptographically secure) and works for Windows UWP targets too, in addition to the code being simpler.

PR #128, opened 2019-11-28
Overhaul list.h to fix various issues

ePirat:
The list.h header had several issues:

- None of the symbols have a upnp prefix, which makes those easy to clash with other global
  names like it already happened for some macOS header defines.
- It defines completely unrelated macros like `bool` and `true`.
- The include guard is incorrectly named
- Contains completely unused hlist

None of these issues would be a huge concern if it was just an internal header, but it is not as it is used in public upnp headers.


mrjimenez:
Nice work, thanks!

whyman:
Thanks! Note this does break the public API, so something to think about when it comes to a release

mrjimenez:
Hi Ian,

It does indeed break the public API, but as far as I can remember, the list API was not used up to "external headers" appeared in scene. And as far as I can tell, you are the only person using it and you are already aware :)

Cheers,
Marcelo.

mrjimenez:
Or should we revert the name we broke?

ePirat:
According to another issue this header was considered "private" and not for public use so I assumed it would be fine, but yes technically this is API-breaking and probably ABI breaking too? (I do not fully understand the macro magic going on)

whyman:
I mean the nicest thing would be to not expose the list stuff at except via wrapped methods, but that would be a lot of work.

As for a release 1.11 or 1.12 would be the sensible thing, then perhaps deprecate 1.10?

ePirat:
I've no idea what the release policy for libupnp is, so I am wondering whats the planned date for the next release?

mrjimenez:
I am just waiting to see if some people will give their opinions on the matter, but if it depended only on me, I would release now. Broken API is serious stuff, it was a good thing that Ian noticed it so soon.

medoc92:
I think that nothing in list.h could ever be considered as part of the libupnp public API. It was just  unfortunately exposed. As far as I can see, only struct list_head is actually needed (e.g. as member of FileInfo)

At the moment, and as far as I can see, only the 1.8 branch is packaged by distributions (and only 1.8.4 which is quite unfortunate because it's broken...)

Would it be a problem to not include this change in 1.8, and then remove the whole of list.h (or at least as much as possible) from newer branches ? libupnp has no business exporting list management APIs.

mrjimenez:
Ok, lets do it.
- 1.8.x was already agreed on not breaking API, so it should be safe.
- 1.10.x will be short lived, no one will package it.
- long live 1.12.x

Notice I did not release 1.12.x, just adjusted the library numbers. This is the time to break API if anyone is interested. What about a window of one week to release?

Regards,
Marcelo.

whyman:
I agree with @medoc92 , I think if we can:

1. Wrap `struct list_head` into a "public" object
2. Allow iteration over the list without using `list_for_each` (wrap it too?)

Then we can make the list header private, which I think is nicest

medoc92:
Oops, I had forgotten the "detail" that the client also needed to manipulate the list. I don't know enough about the use of these lists to really have a competent opinion.

I do feel (and agree with whyman), though, that list.h exposes much too much detail. Would it be possible to have a simple __upnp_list_get_first/__upnp_list_get_next function interface, with appropriate casts when generating the prototypes like currently done in TemplateSource.h for add/remove/replace ? list.h could then probably be reduced to the list_head definition and a handful of externs ? Which would still not be very pretty, but better than the current 850 lines...

But maybe this is stupid again.  I'm not quite sure of how these are used.

medoc92:
I had another quick look at list.h and there is another serious issue with it: It's quite clearly derived from:

https://github.com/torvalds/linux/blob/master/include/linux/list.h

Which is GPL of course.

Using the current list.h would mean that libupnp can only be used by GPL software.  It's a big issue, needing discussion.

mrjimenez:
The original licence of this project is BSD, I don't think we can change that to GPL.

We really don't need the full list.h. I can only think of rewriting the list code. But then the code we need is such a simple and small code that it might not even fit a license.

I agree we need to discuss it.

medoc92:
I definitely agree that it would be preposterous to change the license. I have submitted another pull request to replace the current list.h, I think it works, but I'm not going to fight if nobody likes it, there are many ways to do this.

PR #129, opened 2019-12-01
Remove and replace the list.h file

medoc92:
This is an example of replacing the current list.h 
The list interface is modeled on std::list (a small subset, adapted for c).
I have not tested the implementation with actual extraheaders, I am having trouble building Gerbera
No more macros, but we have to export the list interface routines.
I think that this is much easier to understand than the previous, but this is also a question of taste...

### DIFF
--- a/upnp/Makefile.am
+++ b/upnp/Makefile.am
@@ -26,7 +26,6 @@ upnpinclude_HEADERS = \
 	inc/ExtraHeaders.h \
 	inc/FileInfo.h \
 	inc/list.h \
-	inc/poison.h \
 	inc/StateVarComplete.h \
 	inc/StateVarRequest.h \
 	inc/SubscriptionRequest.h \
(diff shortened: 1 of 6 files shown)


medoc92:
Thanks for merging. It would be nice if someone from Gerbera could give it a try ? As far as I know, Gerbera is the main (or only?) user of the ExtraHeaders function at this point ?

whyman:
Of course I will try it, I agree with @medoc92 that it seems simpler to me, at least.

This does now mean I will end up with 3 versions of this code :man_shrugging:

mrjimenez:
I think we should stick to Jean-François version. Simpler, no license problems and does the job.

medoc92:
@whyman Actually, I had a quick look to the gerbera header.cc, and I don't think that the code needs change, as the xxx_add_to_list_xxx function is still provided by libupnp. It's more a question of checking that it does not crash :)

PR #132, opened 2019-12-05
Add Github Actions based CI

ePirat:


whyman:
Do we need two CI setups?

mrjimenez:
Most likely not. This is not something I have experience with, Marvin, is it safe to remove .travis.yml without breaking the CI integration and the work you did? If that is the case, either tell me or remove it as you please.

ePirat:
Removing the travis file should do no harm at all, as the CI is now using the Github Actions stuff anyway and the travis integration was already broken before. (Which was my main motivation to add the GH Actions stuff instead)

PR #136, opened 2019-12-05
configure.ac: Fix Windows detection

ePirat:


mrjimenez:
"#error" does not work with the windows compiler?

ePirat:
It does not work reliable with some compilers, afaik. No clue which exactly, maybe only old ones we dont care about anymore but better be on the safe side this way…

ePirat:
The more important change here is the `WIN32` to `_WIN32`.

whyman:
Maybe add Windows to the new Github actions CI?

ePirat:
I plan to do this. Adding a Windows build (on Windows using MSVC) is not possible right now though as the MSVC files are completely broken right now.

But I'll probably add cross-build using mingw-w64.

PR #137, opened 2019-12-05
Various compiler warning fixes

ePirat:


ePirat:
Alternatively to the `(void)…` we could make it more explicit by defining a UPNP_UNUSED() macro that does it.

mrjimenez:
> Alternatively to the `(void)…` we could make it more explicit by defining a UPNP_UNUSED() macro that does it.

I like the (void) approach. Never liked that self assignment trick, even less when the compiler started showing warnings.

There is also the #pragma GCC diagnostic approach, but it breaks with some compilers, even with old gcc.

PR #138, opened 2019-12-05
Use stdbool.h instead of BOOL typedef and defines

ePirat:


medoc92:
This will make it impossible to run diffs across the change for no good reason at all (hundreds of changed lines will drown the things you look for). I could understand removing the defines from ixml.h (the only place from where they are exported), but changing the internals is madness.

mrjimenez:
Hi Jean-François,

I understand your point, but the changes are really minor, I believe they are worthy.

But I would not use stdbool.h. Marvin is there a good reason not to use good old '1', '0' and int instead of true, false and bool? After all, this is C :)

Also, stuff like

> if (ixmlNode_compare(attrNode, oldAttrNode) == TRUE) {
> if (Parser_isValidXmlName(qualifiedName) == FALSE) {

Should really be
> if (ixmlNode_compare(attrNode, oldAttrNode)) {
> if (!Parser_isValidXmlName(qualifiedName)) {

Unfortunately I have left this patch to reply later and it is now with minor conflicts, sorry for that.

ePirat:
No reason at all to use bool, happy to use good old 1 and 0.
Should I use `unsigned` as the type for these or just stick with int or something else?

No problem about the conflicts, this was inevitable anyway due to my other PRs…

mrjimenez:
If you have a good reason to use unsigned, go ahead, its your patch.

I would use int because being only three letters it leaves the function prototypes smaller with a better chance of fitting in 80 columns.

ePirat:
No, `int` is absolutely fine for me.

ePirat:
Superseded by #147

PR #162, opened 2020-03-29
add Os/UserAgent information to UpnpActionRequest and UpnpFileInfo

pamapa:
Similar to UpnpDiscovery class add Os member. This implementation is mostly a copy/paste from UpnpDiscovery path, except that i do only look for "USER-AGENT" and not the "SERVER" header entry, as i think the "SERVER" part makes no sense for those "client to server" requests. 

This will give us more flexibility in Gerbera (https://github.com/gerbera/gerbera) when applying quirks.
With this information we can directly use the "USER-AGENT" implementation. For devices where "USER-AGENT" is not good enough, we can use the discovery/CtrlPtIPAddr path.

Works like a charm for VLC + BubbleUPnP

### DIFF
--- a/upnp/inc/ActionRequest.h
+++ b/upnp/inc/ActionRequest.h
@@ -28,6 +28,7 @@
 	EXPAND_CLASS_MEMBER_INT(CLASS, ActionResult, IXML_Document *) \
 	EXPAND_CLASS_MEMBER_INT(CLASS, SoapHeader, IXML_Document *) \
 	EXPAND_CLASS_MEMBER_BUFFER(CLASS, CtrlPtIPAddr, struct sockaddr_storage) \
+	EXPAND_CLASS_MEMBER_STRING(CLASS, Os) \
 
 #include "TemplateInclude.h"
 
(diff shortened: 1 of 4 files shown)


pamapa:
Side note, the USER-AGENT headers are not 100% the same when received by different path:

ActionRequest: 'Android/8.0.0 UPnP/1.0 BubbleUPnP/3.4.4'
Discovery: 'Linux/3.18.91-14843133-QB28034466 UPnP/1.0 BubbleUPnP/3.4.4'

That is coming from the client directly, i do not think this will be a problem, we just a have to consider it when defining USER-AGENT matches.

pamapa:
@mrjimenez 
thanks for merging, does it make sense to backport this to 10.x and 8.x?

mrjimenez:
Hi,

To 1.10.x certainly not, this branch is a mistake, nothing should go in there. Master is 1.12.0 and has been released. Your changes will be in 1.12.1. On this matter, if you feel your are done with your patches, please tell us so that we can release.

To 1.8.x it would make some sense, but then, 1.8.x is supposed to be bug fixes only and these are new features, even though they don't affect old behaviour. I really have no big feelings about that, if the community thinks there is a big reason to backport, I am not against it.

Best regards,
Marcelo.

Issue #106, opened 2019-02-22
[1.8.4] Cannot be built on Windows


emptyVoid:
I'm getting errors while trying to build `release-1.8.4` with `Visual Studio 2017`:
```
Cannot open include file: 'autoconfig.h': No such file or directory
Cannot open source file: '..\..\threadutil\src\FreeList.c': No such file or directory
Cannot open source file: '..\..\threadutil\src\LinkedList.c': No such file or directory
Cannot open source file: '..\..\threadutil\src\ThreadPool.c': No such file or directory
Cannot open source file: '..\..\threadutil\src\TimerThread.c': No such file or directory
Cannot open include file: 'upnpconfig.h': No such file or directory
```

Vollstrecker:
This is an older problem introduced with: https://github.com/mrjimenez/pupnp/commit/c493ccbebe5cc648327fc77b572d6dbfe59598e8

The pregenerated file was removed, and as VS doesn't run autoconf or configure, it isn't genrated again.

l0nley:
Half of the problem solved! Just about 1year passed. Hope 2021 will bring windows builds back :D

emptyVoid:
@mrjimenez, I take it you are not interested in fixing Windows build? Why don't you remove broken Windows configuration and instruction then to make it clear the project no longer supports Windows?

whyman:
@emptyVoid is it same situation with the latest tarball? 

 I imagine the intention is to support windows, but I would assume that @mrjimenez like myself runs *nix exclusively.

Really what we need is a friendly windows developer to help fix up the build :+1:

Vollstrecker:
It's the same with master, except that upnpconfig.h isn't mentioned anymore.

As I stumpled over this while porting aMule to cmake, I could offer to do here the same. I could test on linux and win atm.

emptyVoid:
I've checked `1.12.0` -- headers generated by autotools are missing (autoconfig.h, upnpconfig.h, and config_sample.h). A few compatibility issues with modern Windows SDK and Visual C Standard Library.

I agree CMake-based build system would be great -- one place to define sources for all platforms. Would require some effort to replicate autotools functionality though (like `make dist`).

Vollstrecker:
Why? make dist is meant to create a tarball including such thing like configure pregenerated, so the user doesn't need to run e.g. autogen.

[...]

PR #171, opened 2020-04-17
Fix Windows support when building with autotools under MSYS2.

emptyVoid:
Related to #106.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -683,6 +683,8 @@ else
 	AC_MSG_RESULT([no])
 fi
 
+AM_CONDITIONAL(WIN32, test x"$ac_cv_win32" = xyes)
+
 
 #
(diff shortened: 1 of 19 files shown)


whyman:
Wow nice, can you provide some build instructions too? Would be nice if we could get a windows build into the CI, but I have no idea where to start!

mrjimenez:
It seems like I have somehow missed this pull request. My apologies for that.

emptyVoid:
@whyman, I used `vcpkg` (which has support for `MSYS2`) to build `libupnp` for Windows, creating a custom port and patching some issues in `vcpkg`. IMO this would be an overkill for CI.

Instead I'm looking into `MSYS2`+`MinGW` approach now, and will submit a new PR with fixes for `MinGW` in a bit with the build instructions.

PR #172, opened 2020-05-02
backport mediatomb fix, but place it in httpparser

pamapa:
Original patch:
https://sourceforge.net/p/mediatomb/code/ci/2753e70013636bb5dd4cfc595f9776d368709f04/tree/tombupnp/upnp/src/genlib/net/uri/uri.c?diff=6be2b0d6e512fd60f155acf78b1f7c1725f58571

This patch is needed for "Samsung Smart TV 2012" and older. ~~But it also makes `parse_uri` more robust for everybody:~~
- ~~the schema should only be looked at when we found a schema at all~~
- ~~multiple slashes in URLs are actually allowed and can happen~~

Parse `GET \\cm.xml` to:
- Method = GET
- URL = `\cm.xml` instead of `\\cm.xml`

### DIFF
--- a/upnp/src/genlib/net/http/httpparser.c
+++ b/upnp/src/genlib/net/http/httpparser.c
@@ -1285,6 +1285,11 @@ parser_parse_requestline( http_parser_t * parser )
 
 	hmsg->method = HTTPMETHOD_SIMPLEGET;
 
+	/* remove excessive leading slashes, keep one slash */
+	while ( url_str.length >= 2 && url_str.buf[0] == '/' && url_str.buf[1] == '/' ) {
+		url_str.buf++;
+		url_str.length--;
+	}
 	/* store url */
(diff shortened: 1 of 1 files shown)


mrjimenez:
Hi pamapa,

This patch is breaking one of our unit tests: test_url. Please take a look at the problem and if the test is meaningful after the patch has been applied.

Regards,
Marcelo.

pamapa:
hmm i don't think we can patch pupnp lilke this, i just learnt that `\\localhost` is in fact an absolute URL (protocol-less). I will try to find another way.

pamapa:
In `parse_uri` its already too late as `\\x` must be seen as absolute.

But in httpparser, we see the string like `GET \\cm.xml` and there we can assume must be a relative URL, otherwise it would be very strange, correct?

PR #199, opened 2020-09-30
Fix bad assumptions for OmniOS/Solaris

whorfin:
simple fix
I'm about to give up on gerbera, but while I'm still hanging in there, one more patch...


whyman:
:(

PR #202, opened 2020-11-09
Remove port >= 49152 restriction

lineinthesand:
The media server gerbera is using libupnp. I'm using it with a Yamaha Network Player NP S303 which expects the server port to be 8200. Configuring gerbera using this port is only possible when I remove the requirement that the port be >= 49152.

 This change removes that restriction. I think it is reasonable to assume that if a client requests a certain port number they know what they're doing. If they request port number 0, they get the default minimum port of 49152.

I have an alternative pull request which makes the change configurable.


mrjimenez:
Sorry for taking so long, at first I was waiting so see if someone would complain. I agree with your argument, lets make it default.

mrjimenez:
closes #203

PR #206, opened 2020-12-08
Tests in CMake

Vollstrecker:
Hi,

I finished the running of the tests. If the tests are run from a build that downloaded pthread, the pthread-tests will be run also, coming the a total 309 tests, so testing is not aktivated by default. When the upnp-tests should be run separate it's possible to either switch to upnp or ixml dir and run ctest from there, or give the desired tests as arg to ctest.

init-test succeeds.
list-test succeeds

log-test fails, because the mutex is never initialized but closed in the first lines. I don't know if it is a problem of pthread as it doesn't check for this, or upnp as it shouldn't do that, and I can't imagine how this ever worked in the past.

url-test fails because some windows-functions returns on every hostname. Maybe because I test that in a VM, maybe I forgot a lib the linker doesn't complain about, maybe something changed since the last time compilation on win was possible and noone noticed. I'm not experienced enough with development on win, so someone with more inside has to take a look at this.


whyman:
Just wanted to say this is great work, @Vollstrecker

Vollstrecker:
Not done yet. Beside the samples, I just found a still active pthread4w project and contected them, so maybe we'll get a maintained version of that.

While working on this, I decided to try to get a framework done that is easily to integrate in every cmake-project and intercepts the FindPackage (and maybe more) to get all the deps in. Repo will be created maybe next week when I found a good name.

mrjimenez:
Hi Vollstrecker,

Great job!

About the tests, we have them being run in Github's CI. We have a failing test on MacOS, but we already have figured it out, I just did not have the time to properly fix it. But the tests run and pass on a Linux machine.

Since you are doing a windows build, maybe we should now do a windows build test too on Github CI, what do you say?

Best regards,
Marcelo.

```bash
make  check-TESTS
make[4]: Entering directory '/home/runner/work/pupnp/pupnp/libupnp-1.14.1/_build/sub/upnp'
make[5]: Entering directory '/home/runner/work/pupnp/pupnp/libupnp-1.14.1/_build/sub/upnp'
PASS: test_init
PASS: test_url
PASS: test_log
PASS: test_list
============================================================================
Testsuite summary for libupnp 1.14.1
============================================================================
# TOTAL: 4
# PASS:  4
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
```

Vollstrecker:
If anything doesn't work on linux, I don't commit anything. And for sure you don't get different results, as you run the tests with the original build-system which didn't change at all.

 I will start the Mac-Stuff when I start to do the Mac-Stuff for aMule, so don't expect anything before next summer.

For the windows test, you can include them, if it's a problem with the code, they should pass as you fixed the code. I watch this project, so if anything in build-system needs to be adjusted for this, I hopefully see it, if not just poke me.

For now I pause the work on this. Next step would be the docs, so I need to get Doxygen working, for this I need bison working, for that gettext.... I focus now on that package-manager I mentioned which hopefully ends in making that stuff useable even when upstream doesn't welcome cmake in their source like you did.

mrjimenez:
Before you do your deserved pause, would you mind writing some instructions on how to use cmake to build libupnp for linux and especially for windows users, in README.md?

Vollstrecker:
Sure. Tomorrow.

PR #209, opened 2020-12-16
CMake: Dont postfix debug builds on unix

whyman:


Vollstrecker:
No description? Then a few questions. If I first build a debug version, and later a release one. What happens?

whyman:
It replaces the files at install time? Not sure I understand the question

Vollstrecker:
Corrcet. The debug-enabled lib is replaced with a non-debug version, but the config.cmake files stay as they are. So everyone that uses cmake will be get the debug-flags for compile and then get linker errors, because of the debug-lib doesn't have the functions included.

Your change would be right for libs that don't have different features compiled in, here it just causes troubles.

And I asked for a description to get idea why you think it's good to things that are in there by purpose.

whyman:
Hi @Vollstrecker, firstly thanks for the CMake work, its amazing.

I made this change to bring it into line with behaviour of the existing autotools build system, with the hope that it will replace it at some point.

Linux systems usually don't rename the libraries for debug builds, I have never seen it my time on Linux. Instead the symbols simply get included into the libraries. Or if build by a distribution, the symbols are stripped and shipped separately installable.

If applications are relying on upnp internal debug methods, then that would be very odd indeed, why are those symbols even exported? I see this as more of a bug. If so then they should check the `upnpconfig.h` to ensure that the debug support they need was there, like any other option?

I would imagine that most consumers in the wild are not using CMake, so they just see the libs with the names they expect missing in debug mode.

I may be missing something obvious of course!

Vollstrecker:
Keeping it in line with the autotools sounds good for the first view, but these use pkgconfig, which doesn't differentiate so far. In autotools-scripts it is easy to check if the *d.so there, and then choose what to use.

On windows it's for libs common practice, as they use the lib extension for everything.

The problems with the symbols can be seen in upnpdebug.h, Most of the funcs are inlined when building without debug, and normal when built with. Same for upnpdebug.c, which is just empty in non-debug builds. best would be to remove the ifdef in upnpdebug.c, but that's something @mrjimenez  has to decide.

mrjimenez:
HI,

I have no hard feelings about the way upnpdebug is done. If anyone believes that the current implementation is somewhat problematic, please state your reasons and send a pull request that it will be rationally appreciated.

The current implementation aims to keep the internal upnp code invariant independent of the build mode (debug or not). So, in principle you are able to stuff debug code inside the library without badly affecting code size and speed in non-debug builds, since the compiler would take care of substituting void code by nothing. As a result, debug ifdefs are not required in the library code, and this is a Good Thing (TM), since those debug ifdefs are mainly visual polution.

I agree that other applications should not be using internal upnp debug routines, if that is the case it is definitely buggy and I am in favor of removing the exportation of this "API", but as always I might be missing something.

Regards,
Marcelo.

Vollstrecker:
I don't get it. If you get libs with different functions from different build-modes, I wouldn't call it invariant.

And jep, in eyes it's a good thing to not carry the debug parts through releases, that's why I prefer how it is done now. And sure, the internal routines should be internal only, but they should be there. Given that a developer doesn't get it working like expected, then a blackbox is always bad, the additional debug output is very welcome.

mrjimenez:
> I don't get it. If you get libs with different functions from different build-modes, I wouldn't call it invariant.

The point of view if the written code. The code is invariant, meaning it is the same code for both builds, even if it generates zero executable instructions. The current scheme leaves the ifdefs out of the library code and leave them in the headers, making it more pleasant to read.

As for this patch, I tend to agree with @whyman that in Linux in general packagers don't rename their libraries with a debug version. There are packages with debug symbols that must be installed if you want them to debug or generate readable backtraces. That way the libraries are not recompiled. I don't remember having to link to a library with a different name to debug any application.

In my opinion, the functions in upnpdebug should compile to null code in the real world, even in a postfixed debug enabled library. These functions are supposed to be for the developer's eyes only since they create serious performance and timing issues. The confusion arises from the fact that --enable-debug, which enables debug symbols, also enables this functionality and maybe that is the root of the problem. Should there be separate --enable-debug-symbols / --enable-debug options?

Vollstrecker:
| The point of view is of the written code

k, from the one code for all point of view you're right.

| As for this patch, I tend to agree with @whyman that in Linux in general packagers don't rename their libraries with a debug
| version

Right, too. The usually decide for debug-version strip the symbols to a separate file and good. This is even possible now, even with the release version.

| There are packages with debug symbols that must be installed if you want them to debug or generate readable backtraces

Again right. As the symbols are used for gdb or alike. But all functions are in the lib, and that's the difference to here. If you add DEBUG and UPNP_HAVE_DEBUG as compile-flags to one of the tests, compilation in Release-Mode fails, as the expect the functions from debug (not inlined) in the lib, where they are not found.

It's like when you ship the pkgconfig file with the flags for both, but only ship the release one.

| In my opinion, the functions in upnpdebug should compile to null code in the real world, even in a postfixed debug enabled
| library.

You mean even when debug is requested? Then the easiest way would be to put the inlined functions into a separate header, and force the endusers to include the header with the non-inline version by their own if they really need it. And maybe putting a Warning for packagers to not use debug-build for packaging. But even in this case I would highly recommend to keep the naming difference as it was, as it makes it easier for dev to use the lib if they know what they are doing, without affecting the whole system. If noone uses the debug build, noone will notice the second lib.

Vollstrecker:
Addendum: For sure you never noticed using different lib to link. With autotools you don't create them. wx for example does it when built with cmake, and cmake does the magic to give the right lib to your code.

So no cmake = one lib, no change, no problem
cmake = two libs but you'll never notice, no problem

Vollstrecker:
Addendum 2: If built with cmake and the prog that's is using it uses autotools, it will either search for the lib with no postfix, so no change, or it will use pkgconfig which point it to the non-postfixed release version with the release flags.

So the only difference are some extra bytes on disk for the second lib.

PR #211, opened 2020-12-19
Debug-Stuff now clean

Vollstrecker:
I now created a clean branch and put all the fixes in.
The Problem was a regex that only exports functions starting with Upnp, so instead of prefixing the debug functions I postfixed them now.


mrjimenez:
Hi Vollstrecker,

> Libtool only exports the function when it startes with Upnp

I didn't remember that. I agree, really anoying black magic.

The problem with this patch is that is breaks the library API. That is not something to be done in a hurry, so lets discuss alternatives.

Can we keep the old names? Your last comment on #209 seemed to imply that the problem would be solved if #209 was reverted, is this true?

> Addendum 2: If built with cmake and the prog that's is using it uses autotools, it will either search for the lib with no postfix, so no change, or it will use pkgconfig which point it to the non-postfixed release version with the release flags.
> 
> So the only difference are some extra bytes on disk for the second lib.
 
Please correct me if I am wrong, but I am most inclined to revert #209 than apply this patch that changes the API. @whyman , what do you think?

Best regards,
Marcelo.

whyman:
I created the patch because gerbera failed to find libupnp built via cmake when I tested it, then I found the lib name was different and thought "thats odd".

So I am not overly attached to my patch, but I wonder why it failed if pkgconfig should avoid it.

Vollstrecker:
| I created the patch because gerbera failed to find libupnp built via cmake when I tested it, then I found the lib name was different
| and thought "thats odd".

Why didn't you just tell me that the name in the pc file is incorrect? Beside that, you use with this change you can use find_package(UPNP), and only if that fails, pkgconfig as a fallback. Little hint: better Link to UPNP::UPNP in this case, then you don't have to set includes and that stuff.

| Can we keep the old names? Your last comment on #209 seemed to imply that the problem would be solved if #209 was
| reverted, is this true?

This is almost true. If distros shipped the release-version, it works as always. If they shipped the debug version, they either have to rename (or symlink the lib), or it just will not be found. If they ship Both versions and did only debug before, there will be problems that symbols are not there anymore (except when the build the programs without debug, then nothing happens). But you said one code for all, and this way you have one lib for all, that just depend o compile-flags.

| The problem with this patch is that is breaks the library API

I don't see a problem here. There are renamed functions, but the header maps them to old name, so existing code still compiles without change. Agreed, but a changelogentry that the next release is not binary-compatible should be enough. Most distros (at least in debian I read this on a regular base) have a mechanism to recompile packages against new versions of some libs.

| Can we keep the old names?

As they get always into the lib, there occur errors when building in release-mode, as there and inlined and normal function exist with the same name. It would be possible to decide in the function itself which path to take, but it would be an additional library call instead of the inlined code, which could affect performance.

So for discussing this I summarize

1. With reverting the patch, packagers have to rename the lib, or recompile the depending programms.
2. When not renaming the functions we maybe get performance issues
3. When we keep it as is, we have one code that produces to different libs that are only compatible in one direction.
4. @whyman used the new technics with old tools. (Sorry, but I can't resist)

whyman:
@Vollstrecker Agreed its the "old way" but to be fair, that was the _only way_ until recently so I think I can be forgiven :grin:

Vollstrecker:
I know, and it's my fault that I didn't put the debug stuff in the pkgconfig-file, but this should be ok when we get this stuff here in.

And to be honest, I never checked the find_package stuff outside windows, and guess what: on Linux some have wrong naming and some wrong Paths. There are to many config_file, config-file, CONFIG and combinations of them plus export-name, export_name, name in the docs. All with different meanings.

I'm working on this right now, but I don't think I'll finish today.

Vollstrecker:
Sometimes I'm to blind to see the obvious. The trick is to rename the inlined functions that are not part of the ABI, and just do nothing when included from upnpdebug.c itself.

And find_package should work now as expected, too.

mrjimenez:
Excellent!

Issue #193, opened 2020-07-24
Request for comments: Should an IPV6 enabled libupnp fail initialization if there is no IPV6 available on the chosen interface?


mrjimenez:
Hi folks,

As the title says, I am looking for opinions on how to fix (or not) the current behavior. But first, a little background information.

If you go to the project page today, you will notice that the branch master has a "failing" badge. This is because since we dropped the UpnpInit() method in favor of UpnpInit2(), the mac-os part of the automated tests stopped working. So this problem was always there, but just showed now. That is why I decided to release 1.14.0.

I don't have a Mac, so I had to deal with the virtual Github environment and I created a new branch to do some tests and found out that the mac-os virtual machine did not initialize the IPV6 stack in the en0 (Ethernet) interface. In spite of this, the library did initialize everything correctly until it tryed to create an UDP socket and tried to join IPV6 multicast, which is what triggered the failure.

Fixing the mac-os test failure is not a problem and has indeed been done in the auxiliary branch. But the question that remains is: should the library fail initialization or should it go on with IPV4? I am not the author of the original libupnp IPV6 code, so all I can do is read the code. It seems to me that lots of "ifdefs" were supposed to make the library code smaller, but that is not necessarily the case as the last patch I did on miniserver.c showed. Personally, I believe that if it does not imply a link time failure, being IPV6 enabled is not a problem even if it is not configured. Also, notice that IPV6 might be configured only in some interfaces, which makes things even more messy.

To summarize my thoughts:

1. I believe that the library initialization code should not fail if there is no IPV6 on the desired interface, as long as we can go on with IPV4.
1. I don't believe we need a configure time --enable-ipv6.

Since I can certainly be wrong in both cases, I ask for your comments.

Best regards,
Marcelo.

G10h4ck:
I agree with both your points, still it is quite common to have compile time optional IPv6 support so if it doesn't require too much work to keep it I'd keep the compile time switch

whyman:
The problem is that to get `UpnpInit2` you have to enable IPv6 anyway.

All consumers will now need `UpnpInit2`, so all will require the "optional" IPv6 support.

Therefore it is not actually optional any longer and should be enabled permanently.

Then the library should be made to not fail at runtime with pure v4.

So yes, to both both points from me.

PR #213, opened 2021-01-03
macos: Do not set IPv6 address if not found.

whyman:
The default 0 address gets encoded to :: and therefore fails the empty
check, causing us to try and initalize IPv6 when its not available.

### DIFF
--- a/upnp/src/api/upnpapi.c
+++ b/upnp/src/api/upnpapi.c
@@ -3870,14 +3870,16 @@ int UpnpGetIfInfo(const char *IfName)
 		&v4_netmask,
 		gIF_IPV4_NETMASK,
 		sizeof(gIF_IPV4_NETMASK));
-	inet_ntop(AF_INET6, &v6_addr, gIF_IPV6, sizeof(gIF_IPV6));
-	gIF_IPV6_PREFIX_LENGTH = v6_prefix;
 	gIF_INDEX = if_nametoindex(gIF_NAME);
-	inet_ntop(AF_INET6,
-		&v6ulagua_addr,
-		gIF_IPV6_ULA_GUA,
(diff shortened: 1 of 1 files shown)


whyman:
Fixes the MacOS CI failure :+1:

whyman:
bump :+1:

mrjimenez:
Hi Ian,

I am really sorry I still could not give this issue the necessary attention. Since I had some work done on it, I would like to analyse it with a little more care. Maybe I can do it this weekend.

Best regards,
Marcelo.

whyman:
Hi @mrjimenez 

Happy to wait for you review properly, just wanted to make sure it was not lost in the void.

Cheers!

mrjimenez:
Hi Ian,

Again, sorry for the delay. I just wanted to have the details of the problem in my head before applying your patch.

I understand it addresses the first point of #193, which we both agreed.

Since only you and @G10h4ck have answered, and you have a good point about removing the configure time `--enable-ipv6`, maybe we can go for it now, what do you think?

Regards,
Marcelo.

whyman:
Amusingly I had forgotten entirely about #193 until you mentioned it, and rediscovered this fix by my own investigations.

I think the library should use whichever IP stack is functionally able to support it, be it IPv4 or IPv6 or any combination. We have had for example a report recently that the library fails to start on an "accidentally" pure IPv6 host (https://github.com/gerbera/gerbera/issues/1090 we didnt handle the failure and segfaulted, whoops!) but I would like to enable pure v6 in libupnp at some point.

As for removing the `--enable-ipv6` I think it is a good idea, I would imagine most distributions will have been shipping with that option as Gerbera has required it (due to UpnpInit2) for a number of years now.

That said, it is probably sensible to add at the API layer, rather than build time to enable/disable IPv4/IPv6 support.

PR #219, opened 2021-02-04
Added compatibillity for Debian stable.

Vollstrecker:
I've added some stuff for older cmake-versions. With that debian buster should be also supported #217

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -338,8 +338,40 @@ ENDIF (Solaris)
 if (NOT WIN32)
 	set (THREADS_PREFER_PTHREAD_FLAG TRUE)
 	include (FindThreads)
-	add_library (Threads::Shared ALIAS Threads::Threads)
-	add_library (Threads::Static ALIAS Threads::Threads)
+	if (NOT DOWNLOAD_AND_BUILD_DEPS)
+		if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.18)
+			add_library (Threads::Shared ALIAS Threads::Threads)
+			add_library (Threads::Static ALIAS Threads::Threads)
+		else ()
+			add_library (Threads::Shared INTERFACE IMPORTED)
+			add_library (Threads::Static INTERFACE IMPORTED)
+
+			# The following two blocks replicate the original FindThreads
+			if (THREADS_HAVE_PTHREAD_ARG)
+				set_property (TARGET Threads::Shared PROPERTY
+					INTERFACE_COMPILE_OPTIONS "$<$<COMPILE_LANGUAGE:CUDA>:SHELL:-Xcompiler -pthread>"
+												"$<$<NOT:$<COMPILE_LANGUAGE:CUDA>>:-pthread>"
+				)
+
+				set_property (TARGET Threads::Static PROPERTY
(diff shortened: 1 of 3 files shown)


whyman:
`COMPILE_LANGUAGE:CUDA` ??

Vollstrecker:
Jap. That's how the original FindThreads does it in the old versions. In newer ones it's CUDA or NVIDIA. At least if this is chainloaded in a project where it's needed, noone can blame us, if our test runs before theirs.

https://github.com/Kitware/CMake/blob/05a2ca7f87b9ae73f373e9967fde1ee5210e33af/Modules/FindThreads.cmake#L211

mrjimenez:
Hi Vollstrecker,

I must admit I lack enough CMAKE knowledge to evaluate these patches.

I will commit your changeset, but I would like to leave a request. Would you mind producing a test build based on CMAKE to run along with our current tests? That would make sure things don't break unexpectedly, since it is expected that soon someone other than you, will be changing these files.

The file with the test builds is `.github/workflows/ccpp.yml`.

Best regards,
Marcelo.

Vollstrecker:
> Would you mind producing a test build based on CMAKE to run along with our current tests?

Sure, at least I can try. You'll get a pull-request with some updates for that, as in my fork the tests don't seem to trigger.
Btw. on windows I still get errors for tv_device and tv_combo, both from tv_device.h, I thought this was fixed some time back.

> I must admit I lack enough CMAKE knowledge to evaluate these patches.

Maybe I can give you a little overview, most should be selfexplanatory from the function names. The INTERFACE stuff is the concept that a lib tells someone that links against it what libs and flags to and to tell libs that link to that one, too. Like upnp tells amule to link to ixml and set DEBUG when asked for that flavour. The $<... is more or less if : then : else without retyping the whole command, as the normal if isn't supported within the command.

>  it is expected that soon someone other than you, will be changing these files.

Nice, maybe he can also change all the uppercase commands to more readable, this was needed in earlier versions but not anymore (for a long time now, but old habbits - you know) and the if's and foreach's don't need to repeat the condition when closing.

whyman:
I looked at the windows errors briefly and came to the conclusion they arent designed for windows.

They are in a folder called "linux" after all...

whyman:
Also, I put together CMake based workflow file which may be helpful (Macos, Windows and Linux with CMake & Ninja)

https://github.com/whyman/pupnp/blob/actions-improvements/.github/workflows/ccpp.yml

Vollstrecker:
The file looks good so far, I just don't know what cmake version ubuntu 16.04 has, and if it works.

On line 27 you wrote ONs when it should be ON.

I never give any generator, and never used ninja, but that shouldn't make a difference, but I prefer cmake --build instead of calling
the build-tool by hand.

For the ctest, I prefer to let it print the tests running instead of output-on-failure. As it always could be that I just forgot to add a test (ok not here, atm they are complete)

For the tests you excluded, I'll add an option to CmDaB to decide if the chainloaded test are to be run or not which default to off.

And for the build-errors, as that are redefinition warnings (first in ixml, so internal) and errors about missing parens and semicolon, I'm sure this could be fixed. It's just a pita in my vm, I just hate the concept of VS (sometimes it opens a new file, sometimes it replace the currently opened and that stuff).

PR #221, opened 2021-02-05
Actions improvements

whyman:


Vollstrecker:
Was it planned to replace the autotools tests? I thought it would be an addition.

whyman:
It was on the TODO to add in a sanitizer build, but I havent used it before with CMake.

I think the "log" fixes need some thought too.

Time is lacking as always, though

whyman:
@Vollstrecker the windows tests are failing with `test-upnp-init (Exit code 0xc0000135)` for the shared lib version. Google says that code is DLL not found?

Vollstrecker:
I didn't check by now what the sanitizer actually does. The log, no clue what you're talking about.

For the tests: Known problem https://github.com/pupnp/pupnp/pull/206

mrjimenez:
The macos-10.13 label does not seem to be supported. Also the macos-11.0 label is not currently widely available: https://docs.github.com/en/actions/reference/specifications-for-github-hosted-runners

Maybe we can leave it commented to remember to add it later?

The URL tests on windows seem to be all failing for the same reason, since they are all returning NULL. It can be traced back to resolve_rel_url(), but that function is not debug friendly and I don't have a windows environment here. It could be failing for several different reasons. We could at least add a few debug printf's there, don't you think?

mrjimenez:
..\upnp\src\api\upnpapi.c(69): fatal error C1083: Cannot open include file: 'ifaddrs.h': No such file or directory

#220

mrjimenez:
```bash
3/10 Test  #3: test-upnp-init ...................Exit code 0xc0000135
```
[From stackoverflow](https://stackoverflow.com/questions/11432940/what-does-error-code-0xc0000135-mean-when-starting-a-net-application#:~:text=This%20error%20is%20caused%20when,framework%20before%20running%20the%20application.) it seems that we need to install some kind of .NET framework in the image.

mrjimenez:
Maybe there is some [useful information here](https://docs.github.com/en/actions/guides/building-and-testing-net#starting-with-the-net-workflow-template) on how to set up .NET?

whyman:
I think `0xc0000135` is "library not found".

So I ran:

```
$ cygcheck.exe ./test-upnp-init.exe
C:\Users\Ian\CLionProjects\pupnp\cmake-build-debug\upnp\test\test-upnp-init.exe
  C:\Users\Ian\CLionProjects\pupnp\cmake-build-debug\upnp\test\upnpd.dll
    C:\WINDOWS\system32\WS2_32.dll
      C:\WINDOWS\system32\ntdll.dll
      C:\WINDOWS\system32\RPCRT4.dll
        C:\WINDOWS\system32\KERNELBASE.dll
    C:\WINDOWS\system32\IPHLPAPI.DLL
    C:\Users\Ian\CLionProjects\pupnp\cmake-build-debug\upnp\test\ixmld.dll
      C:\WINDOWS\system32\VCRUNTIME140D.dll
        C:\WINDOWS\system32\ucrtbased.dll
        C:\WINDOWS\system32\KERNEL32.dll
cygcheck: track_down: could not find pthreatd.dll
```

So for some reason it is failing to find the pthreads lib.

I am looking at how we can tell the path where to search.

@Vollstrecker why is the lib called `pthreat`? typo?

Vollstrecker:
Nope, as this is a debug-build the and the pthreads is part of that build, the lib is postfixed, too.

The problem is, that it still checks for the var I set when it was downloading the stuff itself instead of using CmDaB

Change upnp/test/CMakeLists.txt line 3 to: if (pthreads4w_POPULATED) and line 5 to endif() and it should work.

Vollstrecker:
And sorry, I had this problem for weeks, I should've rememberred earlier in this thread. And I have to admit, that I didn't test with preinstalled pthreads, I can maybe tonight or tomorrow. Maybe the can be dumped completely and the whole TEST_ENV var can be build directly in the set_property call.

whyman:
@Vollstrecker I understand why its got the `d` on the end, but shouldnt that make it pthrea**d**d not pthrea**t**d?

This was with the pthreads getting build in the same build using your download thingy.

Ive worked around it for now by telling cmake to dump all the runtime stuff into one directory, that way all the libs are available without messing the PATH.

Vollstrecker:
Sure, fixed.

Vollstrecker:
> Ive worked around it for now by telling cmake to dump all the runtime stuff into one directory, that way all the libs are available
> without messing the PATH.

Sure, that works, too. As the path is only set in the test I see no big difference when it all builds together, but keep in mind, if you've ever installed that all and build with the same install-path again, or you've installed pthreads on it's own, this will be found and nothing will be downloaded. Then the tests will need to know where to find the libs.

Vollstrecker:
Just checked, with my 2 changed lines it works with chainloaded and preinstalled pthreads (at least the error changed from 135 to 139, but I didn't pull anything before that.

If you want to check your solution, don't forget to remove the path-setting, otherwise you won't the difference.

mrjimenez:
Nice, guys!

Vollstrecker:
Ehm, unfortunately this only works with DOWNLOAD_AND_BUILD_DEPS=TRUE. With a preinstalled pthreads not in any system-dir, it still fails. I'll check that tomorrow, and then I'll try to figure out what the sanitizer is doing. Tell me if you need the stuff earlier than I can finish everything.

PR #226, opened 2021-02-08
upnp/src/api/upnpapi.c: don't set gIF_IPV4 if there is no IPv4

ffontaine:
Don't set gIF_IPV4 if no IPv4 is found to keep the default value of '\0' otherwise SSDP will try to register IPv4 multicast with address 0.0.0.0 which will result in a runtime failure


ffontaine:
I'll update this PR shortly as I found another issue possible in the web server

PR #232, opened 2021-02-14
Collected Work

Vollstrecker:
Hi,

I've never looked at the numbers, just if they are there, but this should fix it. In addtion I droped the all uppcase from legacy cmake.
For the tests, they now should work with every situiation. I splitted the building in Release and Debug as the problem only exists in Release builds, so everything should work, or better the release part should fail, if it's not a local problem on my system.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -4,12 +4,13 @@ on: [push, pull_request]
 
 jobs:
   linux:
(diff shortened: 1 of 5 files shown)


Vollstrecker:
Ok, they don't fail. I have to check with a different vm, Maybe the problem is at this.

whyman:
The lowercase cmake looks 10x nicer!

PR #233, opened 2021-02-16
Added googletest and initial testsuites

ingo-h:
Download googletest from github with cmake into the project to simply
keep it up to date and to keep the project lean.

Just install with `cmake .` in `pupnp/`. It will download googletest to `pupnp/googletest-build`, `pupnp/googletest-src` and `pupnp/googletest-download` and build it:

    pupnp$ cmake .  # there is a dot
    pupnp$ make     # create the library
    pupnp$ cd googletest-build
    googletest-build$ make   # create googletest

This will create the different static linked programs in `pupnp/lib`. Use them to link your testsuites. Have a look at `pupnp/gtest/run_test`. It's a bash script. At first you should check the test environment by running the test_template.cpp. It must end without failures and output "PASSED":

    gtest$ ./run_test test_template.cpp

Because the tests are compiled you can just run them again if you don't have made changes on the tests or on the library:

    gtest$ ./test_template.a

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -475,3 +475,35 @@ CONFIGURE_FILE (${CMAKE_CURRENT_SOURCE_DIR}/libupnp.pc.in ${CMAKE_CURRENT_BINARY
 INSTALL (FILES ${CMAKE_CURRENT_BINARY_DIR}/libupnp.pc
 	DESTINATION ${CMAKE_INSTALL_LIBDIR}/pkgconfig
 )
+
+# Download and unpack googletest at configure time
+configure_file(CMakeLists.txt.in googletest-download/CMakeLists.txt)
+execute_process(COMMAND ${CMAKE_COMMAND} -G "${CMAKE_GENERATOR}" .
+  RESULT_VARIABLE result
+  WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/googletest-download )
+if(result)
+  message(FATAL_ERROR "CMake step for googletest failed: ${result}")
+endif()
+execute_process(COMMAND ${CMAKE_COMMAND} --build .
+  RESULT_VARIABLE result
+  WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/googletest-download )
+if(result)
+  message(FATAL_ERROR "Build step for googletest failed: ${result}")
+endif()
+
+# Prevent overriding the parent project's compiler/linker
+# settings on Windows
+set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
+
+# Add googletest directly to our build. This defines
+# the gtest and gtest_main targets.
+add_subdirectory(${CMAKE_CURRENT_BINARY_DIR}/googletest-src
+                 ${CMAKE_CURRENT_BINARY_DIR}/googletest-build
+                 EXCLUDE_FROM_ALL)
+
+# The gtest/gtest_main targets carry header search path
+# dependencies automatically when using CMake 2.8.11 or
+# later. Otherwise we have to add them here ourselves.
+if (CMAKE_VERSION VERSION_LESS 2.8.11)
+  include_directories("${gtest_SOURCE_DIR}/include")
+endif()
--- a/CMakeLists.txt.in
+++ b/CMakeLists.txt.in
@@ -1,0 +1,15 @@
+cmake_minimum_required(VERSION 2.8.12)
+
+project(googletest-download NONE)
+
+include(ExternalProject)
+ExternalProject_Add(googletest
+  GIT_REPOSITORY    https://github.com/google/googletest.git
+  GIT_TAG           master
+  SOURCE_DIR        "${CMAKE_CURRENT_BINARY_DIR}/googletest-src"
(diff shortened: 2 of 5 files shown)


Vollstrecker:
I'm pretty sure this will work without external project, without a generated additional CMakeLists.txt, without hardcoded compiler (and on more linux), without static-linking only, without bash and with inclusion in the main progress.

I'll do some tests tomorrow.

ingo-h:
Except a very small `CMakeLists.txt.in` and of course the testsuites itself there is nothing in included  in the main progress.

Vollstrecker:
Exactly, but it should be.

Vollstrecker:
And a break from multiplattform cmake into bash over statically linked executable and back ist just a nogo. Beside the fact that you took the uncomfortable way of downloading gtest and don't check if it's maybe existing on the system.

ingo-h:
You are right with cmake and bash. The test environment is far from being optimal. It is the first step and should be understood more as proof of concept and should of course be improved. But it is not production code. It is a helper to make leaner and more stable production code with mulithreading and dynamic linked libraries. Of course you can set these requirements for the test environment as well, but this is the wrong approach. Other criteria apply, such as ease of use and quick but not dirty made tests. This is given with statically linked libraries. Its no problem to make them dynamically linkable but with some more effort, less flexible and more error prone. The test code does not fulfill any productive tasks, except for productive software development.

Looking for another googletest installation doesn't help much. It has a reason to compile it together with the project. There are some possibilities to install googletest. At [incorporating-into-an-existing-cmake-project](https://github.com/google/googletest/blob/master/googletest/README.md#incorporating-into-an-existing-cmake-project) you will find 4 different ways to install it. The last one is noted as follows:
>Use CMake to download GoogleTest as part of the build's configure step. This is just a little more complex, but doesn't have the limitations of the other methods.

I prefer the best method.

Vollstrecker:
No offense, I appreciate that you improve the tests. As I gave you the CmDaB-line, I told you that googles approach is bad. Admitted, this was a quick-and-dirty addition, but if it's not working you just can tell me, I just took the time to work on other things.

> I prefer the best method

Me too. Maybe I should clarify why this approach is not the best method. And yes maybe there points in the list that are proof-of-concept related, but not all.

- This download happens unconditionally. No check if gtest is installed on your system, no check if tests are to be build.
- ExternalProject usually does it's work at build-time. That's why the additional CMakeLists.txt needs to be generated. Then
  a new cmake process is spawned which repeats all the checks that already happened (i.e. the compiler) with it's own cache
  and own functions that are not accessible from the main project?
- There is a separate build-step just for gtest needed. If the tests are to be run, then everything to do that should be included
  in the build, if no test should be run, nothing should be built.
- For the static linking, sure this is no productive code, but you are testing libs that are supposed to be used productive. As I
  unified debug and release functions into one lib, static linking worked, but the dynamic libs just didn't export their symbols
  (thanks to libtool-voodoo). Catching this is with only static linking impossible, so both needs to be done.
- The bash is clear. This tests only run on if gcc and pthread are available, and if the lib was built with IPv6-support. I don't know
  about Mac, but Win is absolutely out.

So what's the best method? I really don't know. But not using existing infrastructure, going to linux-only as we just got back multiplatform support and let the user do extra steps when not needed is definitely not.

I don't understand the concept of gtest completely, and as I forgot so many things over time, I just have a rough guess what you tests are doing, and I have the feeling that you have the same problems with cmake. So I again offer you (I will do anyway) to rework the cmake parts, so you can concentrate on clean tests.

I fully understand that you needed to get just a way to get your tests running, and the bad docs from google are not your fault.

ingo-h:
@Vollstrecker I completely agree with you. That are all limitations from the proof that should be optimized.
>So I again offer you (I will do anyway) to rework the cmake parts, so you can concentrate on clean tests.

Thank you. That's what I mean by collaboration. I'm not familiar with cmake and builds. But wait a bit. I'm just working on a big issue with mocking. I have [mocked tests running but they all finish with a memory leak](https://stackoverflow.com/q/66260511/5014688). That must be fixed before going on.

>I fully understand that you needed to get just a way to get your tests running

That's right.

mrjimenez:
So, is the patch good to go in, or should we wait? I have no problem with work in progress as long as there is no breakage.

Please do consider that we are about to release 1.14.2.

Vollstrecker:
I would suggest ti resync and then only include the 2 test-files (when the memleak is gone). I then throw the other stuff around it. I'm not sure if get ready today, but I'm heading to finish-line with it.

ingo-h:
As far as I have seen the memory leak seems not to be a big problem and can be suppressed. But I'm still working on mocking. In my opinion it makes little sense to add the testing now as long as mocking of the interfaces is not working. The problem is to overwrite the low level ifaddr structure getting with `getifaddrs(struct ifaddrs **ifap)` under test conditions to fake different network interfaces to the library. So I suggest to wait with integration of the tests. It's still too unsafe for a release.

mrjimenez:
Ok then. Please tell me when it is ready.

ingo-h:
@mrjimenez I'm ready now to go for testing and mocking. I have managed to fix the memory leak issue and to mock hardware interfaces, e.g. network interfaces.

PR #234, opened 2021-02-17
Fix crash in UpnpGetIfInfo

jbroadus:
Per getifaddrs documentation, the ifa_addr field of an ifaddrs structure
can be null. In a real world example, an entry may be provided for the
non-existent hardware address of a tunnel device. This behavior was
observed with the netlink based getifaddrs implementation in glibc.

### DIFF
--- a/upnp/src/api/upnpapi.c
+++ b/upnp/src/api/upnpapi.c
@@ -3771,6 +3771,10 @@ int UpnpGetIfInfo(const char *IfName)
 			(!(ifa->ifa_flags & IFF_MULTICAST))) {
 			continue;
 		}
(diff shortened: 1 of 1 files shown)


mrjimenez:
Thanks!

jbroadus:
Thanks for merging, but with PR #235, this one is redundant.

mrjimenez:
Oh, I see. I'll fix the redundancy, but your contribution is registered and your patches are welcome.

Regards,
Marcelo.

jbroadus:
> Oh, I see. I'll fix the redundancy, but your contribution is registered and your patches are welcome.
> 
> Regards,
> Marcelo.

Thank you, kind sir.

PR #238, opened 2021-02-24
Added gtest detection, so @ingo-h can see how to include his tests.

Vollstrecker:
I didn't include the tests, so he can finish his stuff. I wanted to PR this to master originally, but it seems the other changes went in 1.14 only, so maybe this should get merged there, too and then all together pulled into master.


Vollstrecker:
Strange, the code for version detection hasn't been touched.

mrjimenez:
You are aware that this patch does not pass any of the windows tests, right? Can you please confirm that it is supposed to be merged anyway? I am putting new stuff into master, if you say is the case, we can put it into 1.14.x.

Somehow I don't think this is a good idea to put this into 1.14.2, since it fails tests and I should have released it yesterday. I am having issues with my computer right now.

Vollstrecker:
> You are aware that this patch does not pass any of the windows tests, right?

Jep, my vm needed a new setup with updates, reboots and all that stuff and gtest seems to make some things different on win and I'm still waiting to get a look at this.

> Can you please confirm that it is supposed to be merged anyway?

Not with failing tests of course.

> I am putting new stuff into master, if you say is the case, we can put it into 1.14.x.

Maybe I did something wrong while rebasing, while creating this one, I was told it's unmergeable. For 1.14.x it's only a good idea if the tests for gtest get ready in time, what I don't believe.

Vollstrecker:
Bu the last one needs to go into 1.14 aswell, otherwise libupnp will always report x.0.0. as version.

PR #242, opened 2021-03-04
Expose ExtraHeaderList in the ActionRequest interface

AlaricSenat:
This PR exposes unrecognized headers in the `UpnpActionRequest` interface
Api users might want to process some http headers that are not recognized by the pupnp http server. A similar behavior is already implemented for the upnp http io callbacks in the `FileInfo` interface.

### DIFF
--- a/upnp/generator/generator.c
+++ b/upnp/generator/generator.c
@@ -966,7 +966,7 @@ int main(int argc, const char *argv[])
 		if (!fp) {
 			continue;
 		}
-		printf("Writting %s ... ", c[i].header);
+		printf("Writing %s ... ", c[i].header);
 		write_header(fp, c + i);
 		fclose(fp);
 		printf("done!\n");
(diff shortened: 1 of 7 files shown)


mrjimenez:
I must say that since the new list code has added, I have not seen such a large modification in this library.

I must also say that even though I have reviewed the patch as carefully as I can, I am unable to test it. If it is not asking too much, maybe in a future patch you might consider adding some unit tests in the test programs?

Great work!

Regards,
Marcelo.

AlaricSenat:
Thanks!

It seems that the HTTP parser isn't covered at all by the unit tests.
I'd gladly add a test or two to test the code I added, even if most of it is just moving around already existing functions. But then it'll feel weird cause i'll only test just a tiny part of the HTTP parser code... 
Partial test coverage is better than none I guess !

mrjimenez:
> Partial test coverage is better than none I guess !

You bet! :laughing:

PR #254, opened 2021-03-13
gtest simplification

Vollstrecker:
Hi,

I just took the time and added googletest to CmDaB. So now the second generated CMakeLists isn't needed, and everything runs smooth in on process.

For now I left the compile.sh and run_test scripts in, but as they have hardcoded output-path's and path's where to look for gtest, and are not run at all by now, I hardly doubt that they make any sense in here.

I also ignored the subdirs of the gtest dir, as they seem to me not like tests and more like examples for gtest. In this case I also don't see a reason for keeping them. Except for the tools dir, as they are used in the upnpapi tests.

For now this all is excluded on Win, as the test unconditionally includes <ifaddrs.h>, <arpa/inet.h> and <net/if.h> which just don't exist.

And for the failing tests, I don't know what is failing there, but it's in the containers. As long as one Ubuntu-Build succeeds, everything is fine.


Vollstrecker:
And I just checked, on Mac the tests don't work, too. But there seems to be a problem in the test-scenario as build runs fine.

Vollstrecker:
For win did a short check, if removing the includes and using UpnpInet.h instead, it goes a bit further, but @ingo-h used inet_aton and all that stuff. Yep, it's nice the be able to mock some things, but a multiplattform lib should also have multiplatfrom tests.

ingo-h:
With some subdirs it was the idea to show how it works or how it is made so that you can understand it better. If it is not wanted it's no problem to omit them. Having the tests running for the other platforms is one of the steps in the near future. I will make a ToDo list. The idea is for example to run the MS Windows tests together with the Linux tests. As already told in the past, the compile.sh are really not the final solution. What will be used? Cmake scripts? If one can be given as example I would use it.

Vollstrecker:
Take a look at gtest/CMakeLists.txt, there you should be able to reproduce as many tests as you want. Maybe I'll pack that in a wrapper function when more test come in.

For the subdirs, maybe @mrjimenez find's it interesting or usefull, if not we need to know what can be removed. I guess everything except tools/

mrjimenez:
Guys, I leave it all up to you. Feel free to experiment. I just ask for a little documentation so that we all know how to test and benefit from it.

Regards,
Marcelo.

ingo-h:
@Vollstrecker OK, I will use that CMakeLists.txt with a little learning curve. I will remove the extra subdirs and put a documentation into the README and on stack overflow.

Issue #248, opened 2021-03-10
Using classes and objects in the library code


ingo-h:
The UPnP protocol specification recommends to write code in an object oriented way. So I ask if it is possible to also use classes and objects for the code of the library. I don't think about to rewrite the whole library at once. There is to much know how into the code. But it may be possible to pick up subroutines/functions and rewrite them as classes. Together with Unit testing we could ensure that the behavior does not change. So we could do a step by step refactoring to use OOP. What do you think?

mrjimenez:
Hi Ingo,

At this point, I don't see an advantage to do that. I told you before that I have though about rewriting it in C++, but then I thought for some time and didn't go ahead. The main reason (besides the lack of time) being that C++ usually generates a greater amount of code, while in C you have the opposite. Since the library is used in embedded systems, where memory footprint might be an issue, in the end I have chosen not to go through that path.

Your argument about unit test is good, but unit testing is also possible without C++, though certainly not as beautiful and easy.

Also, make no mistake: C++ is not C. I have dealt with both and it is much harder to understand what is really going on on C++. That extra not always fully understandable complexity is a little hard to use on your advantage on smaller and simpler programs.

As I said on the other thread, I imagine it would be a very nice code to write if properly modeled. But I also imagine it would be a total rewrite of the library.

I really do not want to discourage you, because I believe that people see things in different ways and have different experiences, and you might be seeing something amazing that I am not seeing. So I would like to add that if you still believe OOP would fit some subsystem of libupnp, definitely go for it. Worst case scenario you have a friendly fork of the project. And at the same time, best case scenario it will eventually become the new maintained one. Open source is all about that.

If you decide to go that way, there are some places where you can start. A long time ago I decided to create some C-like classes. You can have a look at them in `upnp/inc/Upnp*.h`. These files are mostly automatically generated by another program called `generator.c`. Not a long time ago, those classes were preprocessor generated, but the result was kinda messy and now the old templates are gone. These classes are far from incomplete, they are just the basic functionality I did in order to avoid rewriting the same code over and over again. Maybe you could rewrite one of these classes in C++ and see what the generated code looks like and behave in terms of size and speed. If you can show that there is an advantage, I am sure that people will support it.

Best regards,
Marcelo.

Vollstrecker:
[...]

PR #255, opened 2021-03-14
Reentrant version of libupnp

mrjimenez:
Uses a library handle to isolate global variables.

Extremely experimental.


mrjimenez:
Is there anyone against committing this? There's some API breakage, and there is a lot to be tested.

Vollstrecker:
Usually I would say "You do it right, no matter what you do", but I would keep it in a branch until it compiles.

whyman:
LGTM, breaks the tests though :smiley: 

I would maybe consider `UpnpLib` -> `UpnpHandle`?

In terms of usage, I was imagining `UpnpInit2` to return the handle, rather than having to create one and pass it in (like e.g. `curl_easy_init()`)

mrjimenez:
Hi @whyman ,

> LGTM, breaks the tests though 
The broken tests are the ones I was unable to do here. The sanitizer build, which uses autotools is passing the tests.

> I would maybe consider `UpnpLib` -> `UpnpHandle`?

Not a problem, just some repetitive work, but seems ok. I have called it "Library Handle" in the docs, I have chosen the lib part, and you have chosen the handle part :laughing: 

> In terms of usage, I was imagining `UpnpInit2` (or 3) to return the handle, rather than having to create one and pass it in (like e.g. `curl_easy_init()`)

This is certainly better, I agree. Breaks the "return error" pattern of the library, because it would return a pointer and we would have to consider returning NULL an error. Today, returning a zero (`UPNP_E_SUCCESS`) is ok. But I like your suggestion. Maybe we could even return to `UpnpInit`, since it won't go anywhere without the handle.

Hi @Vollstrecker ,

> Usually I would say "You do it right, no matter what you do", but I would keep it in a branch until it compiles.

It does compile and passes the rather simple tests in the autotools package.

There are also your suggestions in #248 to take into consideration.

mrjimenez:
Well, now it does not compile. Is there any way to convince the C++ compiler to accept it or should we reverse the last change?

Vollstrecker:
That's what I meant. I don't think it's cmake's fault as it just does what it is told to do. I think more that libtool is doing some voodoo again and decides to use the c-ompiler.

I didn't expect that you merge this one, as the tests failed. I just wanted to show you what problems are reported if forced to cpp-mode. I dig deeper into letting libtool show me the used command instead of just CC, to find out what happens.

Btw. The set_source.. would really be obsolete if *.cpp is used instead of *.c.

wooky:
If you want to use the C++ compiler to compile C files, you can always surround it with
```c
#ifdef __cplusplus
extern "C" {
#endif
```
and
```c
#ifdef __cplusplus
}
#endif
```
Alternatively, we'll just have to find all the C++ errors and resolve them, one by one.

mrjimenez:
@wooky , 

Well, this was not my intention right now, maybe a little more ahead when the tests with CMake were passing.

I actually got ahead of myself and did not properly check what @Vollstrecker was doing. I though he had figured out a way to make the tests work, my fault.

But your suggestion is correct, if/when we start using C++ we will probably need to force some C code into C++.

> I didn't expect that you merge this one, as the tests failed. I just wanted to show you what problems are reported if forced to cpp-mode. I dig deeper into letting libtool show me the used command instead of just CC, to find out what happens.

@Vollstrecker , sorry for that, I will revert it.

> Btw. The set_source.. would really be obsolete if *.cpp is used instead of *.c.

If I understand you correctly, you are saying that we should rename what we want to use as C++. I agree with that. But this is the gradual change I referred to before.

mrjimenez:
Ok, just reverted the last two changesets. @Vollstrecker , is it possible to temporarily disable the gtest in this branch? I believe the tests would pass then.

Vollstrecker:
Easiest would be to comment Line 91 of Toplevel CMakeLists. Cleaner the whole block from 91-100

Vollstrecker:
I just tried with disabled gtest, but as the error comes from UpnpActionComplete.c, this doesn't change that much.

https://github.com/Vollstrecker/pupnp/actions/runs/656768691

>If I understand you correctly, you are saying that we should rename what we want to use as C++. I agree with that. But this is the gradual change I referred to before.

Not, if you just create the file and then move the parts when done. I mean you create the class, that uses the old functions as they are. Let's take UpnpActionComplete.c, the  upnpHandle can work with the old one and the new one goes into UpnpActionComplete.cpp. Maybe with a prefix to see the difference, or with a namespace protecting it.

I know, confusing to read. I could try to just write the code as example, but for this week (originally 6 weeks ago) I planned to focus on cryptopp.

mrjimenez:
```bash
...
[122/135] Building C object upnp/sample/CMakeFiles/tv_ctrlpt.dir/common/sample_util.c.o
[123/135] Building C object upnp/sample/CMakeFiles/tv_ctrlpt.dir/common/tv_ctrlpt.c.o
[124/135] Linking C static library upnp/libupnp.a
[125/135] Building C object upnp/sample/CMakeFiles/tv_ctrlpt.dir/linux/tv_ctrlpt_main.c.o
[126/135] Building C object upnp/sample/CMakeFiles/tv_combo.dir/common/sample_util.c.o
[127/135] Building C object upnp/sample/CMakeFiles/tv_combo.dir/common/tv_ctrlpt.c.o
[128/135] Linking C executable bin/tv_ctrlpt
[129/135] Building C object upnp/sample/CMakeFiles/tv_combo.dir/common/tv_device.c.o
FAILED: : && /usr/bin/cc -g  upnp/sample/CMakeFiles/tv_ctrlpt.dir/common/sample_util.c.o upnp/sample/CMakeFiles/tv_ctrlpt.dir/common/tv_ctrlpt.c.o upnp/sample/CMakeFiles/tv_ctrlpt.dir/linux/tv_ctrlpt_main.c.o -o bin/tv_ctrlpt  -Wl,-rpath,/home/runner/work/pupnp/pupnp/build/upnp:/home/runner/work/pupnp/pupnp/build/ixml:  upnp/libupnp.so.18.0.0  ixml/libixml.so.11.0.0  -pthread && :
upnp/libupnp.so.18.0.0: undefined reference to `UpnpLib_set_LOCAL_PORT_V4'
upnp/libupnp.so.18.0.0: undefined reference to `UpnpLib_getnc_gDocumentRootDir'
upnp/libupnp.so.18.0.0: undefined reference to `UpnpLib_getnc_gRecvThreadPool'
upnp/libupnp.so.18.0.0: undefined reference to `UpnpLib_get_gIF_IPV6_ULA_GUA_Length'
...
upnp/libupnp.so.18.0.0: undefined reference to `UpnpLib_set_LOCAL_PORT_V6'
upnp/libupnp.so.18.0.0: undefined reference to `UpnpLib_set_bWebServerState'
collect2: error: ld returned 1 exit status
ninja: build stopped: subcommand failed.
Error: Process completed with exit code 1.
```

Any ideas?

mrjimenez:
The failure now is due to the inclusion of openssl/ssl.h, which must be conditional to UPNP_ENABLE_OPEN_SSL. That requires some logic in generator.c to write UpnpLib.{c,h} properly.

mrjimenez:
I don't understand what caused the error on windows builds:
```bash
[305/447] Building C object upnp\CMakeFiles\upnp_shared.dir\src\api\UpnpStateVarRequest.c.obj
[306/447] Building C object upnp\CMakeFiles\upnp_shared.dir\src\api\upnpdebug.c.obj
..\upnp\src\api\upnpdebug.c(89): warning C4996: 'fopen': This function or variable may be unsafe. Consider using fopen_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.
C:\Program Files (x86)\Windows Kits\10\include\10.0.17763.0\ucrt\stdio.h(208): note: see declaration of 'fopen'
..\upnp\src\api\upnpdebug.c(93): warning C4996: 'strerror': This function or variable may be unsafe. Consider using strerror_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.
C:\Program Files (x86)\Windows Kits\10\include\10.0.17763.0\ucrt\string.h(182): note: see declaration of 'strerror'
..\upnp\src\api\upnpdebug.c(139): warning C4996: 'strdup': The POSIX name for this item is deprecated. Instead, use the ISO C and C++ conformant name: _strdup. See online help for details.
C:\Program Files (x86)\Windows Kits\10\include\10.0.17763.0\ucrt\string.h(536): note: see declaration of 'strdup'
..\upnp\src\api\upnpdebug.c(215): warning C4996: 'localtime': This function or variable may be unsafe. Consider using localtime_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.
C:\Program Files (x86)\Windows Kits\10\include\10.0.17763.0\ucrt\time.h(506): note: see declaration of 'localtime'
ninja: build stopped: subcommand failed.
Error: Process completed with exit code 1.
```

mrjimenez:
Now it just remains the windows stuff.

whyman:
Looks like the relevant windows errror is:

``` ..\upnp\src\api\upnpapi.c(3615): error C2037: left of 'gIF_NAME' specifies undefined struct/union 's_UpnpLib'
..\upnp\src\api\upnpapi.c(3619): error C2037: left of 'gIF_NAME' specifies undefined struct/union 's_UpnpLib'
..\upnp\src\api\upnpapi.c(3619): warning C4090: 'function': different 'const' qualifiers
..\upnp\src\api\upnpapi.c(3619): error C2198: 'strncpy': too few arguments for call
..\upnp\src\api\upnpapi.c(3637): error C2037: left of 'gIF_NAME' specifies undefined struct/union 's_UpnpLib'
..\upnp\src\api\upnpapi.c(3638): warning C4133: 'function': incompatible types - from 'PWCHAR' to 'char *'
..\upnp\src\api\upnpapi.c(3639): error C2037: left of 'gIF_NAME' specifies undefined struct/union 's_UpnpLib'
..\upnp\src\api\upnpapi.c(3639): error C2198: 'wcstombs': too few arguments for call
..\upnp\src\api\upnpapi.c(3653): error C2037: left of 'gIF_NAME' specifies undefined struct/union 's_UpnpLib'
..\upnp\src\api\upnpapi.c(3655): error C2037: left of 'gIF_NAME' specifies undefined struct/union 's_UpnpLib'
..\upnp\src\api\upnpapi.c(3655): error C2198: 'strncmp': too few arguments for call
..\upnp\src\api\upnpapi.c(3704): error C2037: left of 'gIF_INDEX' specifies undefined struct/union 's_UpnpLib'
..\upnp\src\api\upnpapi.c(3719): error C2037: left of 'gIF_IPV4' specifies undefined struct/union 's_UpnpLib'
..\upnp\src\api\upnpapi.c(3719): error C2198: 'inet_ntop': too few arguments for call
..\upnp\src\api\upnpapi.c(3720): error C2037: left of 'gIF_IPV6' specifies undefined struct/union 's_UpnpLib'
..\upnp\src\api\upnpapi.c(3720): error C2198: 'inet_ntop': too few arguments for call
```

Missing include perhaps?

Vollstrecker:
I remember the gIF* stuff from the errors in the gtests. Somehow related to UpnpInet.h, but I didn't investigate that further as inet_aton stopped the build after changing the included. I think the error in the win-builds of the tools as they failed building looked somehow similar.

mrjimenez:
Hi @whyman 
 
> Missing include perhaps?

No, more like missing windows here :laughing: Since I got no complaint for what was inside `#ifdef WIN32`, that code has not been fixed. I'll give it a go later, it is just manual work.

mrjimenez:
I don't know why the tests fails on windows. They work on Mac and Linux. Maybe someone can debug it or get a backtrace?

Vollstrecker (commented):
I'm pretty sure you want cd build/bin

Vollstrecker (commented):
Sorry, missed that one. You want test-upnp-init

mrjimenez:
```bash
Run cd build/bin
  cd build/bin
  test-upnp-init
  type test_init.log
  ctest --output-on-failure
  shell: C:\Program Files\PowerShell\7\pwsh.EXE -command ". '{0}'"
test-upnp-init: D:\a\_temp\10150e3e-81c4-48bc-8e90-a557281443a3.ps1:3
Line |
   3 |  test-upnp-init
     |  ~~~~~~~~~~~~~~
     | The term 'test-upnp-init' is not recognized as a name of a cmdlet, function, script file, or
     | executable program. Check the spelling of the name, or if a path was included, verify that the path is
     | correct and try again.

Error: Process completed with exit code 1.
```

Vollstrecker:
k, ctest shouldn't work in build/bin. at least it's at least undefined. And as this is on Win, maybe build\bin is the right thing, which will break this on linux for sure. I'm going to play with this a bit tomorrow (just getting you're commands to run, not trying to understand what you do with them).

mrjimenez:
Ok, thanks!

Vollstrecker:
Just gave it a short try with dir before and after the test. it seems that the ...ps1:3 line comes from the test and everything after that just doesn't get executed. So I'm not able to help you anymore with that. I just can suggest the vm's to run the debugger and view it live.

mrjimenez:
> What I can see is upnpapi.c line 434 the ithread_mutex_unlock call results in a problem because of a nullptr in mutex->kind.

That information was gold. Thanks!

PR #262, opened 2021-03-19
Some improvements

Vollstrecker:
The last commit first. As it seems that @whyman  blocks anything that changes the libname to something different than upnp, I now changed the output_name on windows. Before we got upnp.dll, now we get libupnp.dll which is the same as on the other plattforms. I still thik this is the worst solution (beside building only static on win), it's better to have a bad one then nothing. I for myself don't care, I use UPNP::shared and never have to know how the real lib is called. For any other projects that maybe have to adjust their self-fiddled scripts: I don't care.

The others just move some parts from to toplevel dir into a cmake subdir similar to m4/.

As @ingo-h  wanted, I made it easier to add tests (both unit and getest). And no, this functions are not documented, beside the find TestLibs function which is recursivly calling itself until it's finished, everything in there should be understandable for anyone that knows a little cmake. All others should consider this a black-box and just use the calls that should be self-explanatory in /upnp/test/CMakeLists.txt and gtest/CMakeLists.txt.

I didn't take into account that unittests could need addtional includes, too. As I don't expect new stuff there, this should be OK. If I'm wrong, it's easy to extend. For gtests I didn't consider that they could need only gtest, they always get gmock. As ingo stated everything should be mockable, I don't expect tests without mock. If I'm wrong here: gmock links against gtest, so it should work anyways. As long as this are tests and not (memory-)benchmarks, this should be OK also.


whyman:
I may be missing it but it looks to me this way we lose the GTest CTest integration via `gtest_discover_tests`. If a test executable contains multiple (g)tests its nice to have each test status reported separately.

Obviously the tests are standalone (1 test -> 1 executable) currently, but they will likely not be in the future.

re: Blocking, I do not block anything, merely repeat my opinion :smile:

whyman:
Yep, I did miss it, there is `gtest_add_tests` instead :D

Vollstrecker:
discover_test parses the output of the binary for the tests, add_tests parses the source. With discover it wouldn't be possible to set env-vars for test that are not created at that time.

Vollstrecker:
And we maybe have to overwork the debugging of the windows-test.

Vollstrecker:
And yes, the gtest dir isn't included. I don't know if it's just failing on my clone or everywhere. I maybe have to cleanup again.

mrjimenez:
> And we maybe have to overwork the debugging of the windows-test.

On that point, I don't understand the failure, all the test seems to have run until the end.

Vollstrecker:
If you set the value, you have to mask the ;
If you return it from a function and set then, you have to double-mask.
Sometimes it's that easy.

PR #263, opened 2021-03-19
Use C++ std::string while preserving the old C interface

mrjimenez:


Vollstrecker:
May I suggest to drop Ninja? It's liked in the wild because of it's speed and mutlithreading abillities, but readability of failed builds doesn't benefit from that.

mrjimenez:
@Vollstrecker ,

You are the CMake ninja, please be my guest. :laughing:

Seriously, if you think it is more readable, go ahead, I am really new to this CMake thing and still getting used to it.

PR #264, opened 2021-03-19
Drop Ninja

Vollstrecker:
... and VSwhere (whatever that is)
and Choco (doesn't know either)

cmake-ninja's don't know the yml, so hopefully this works. At least the syntax check seems to pass.


Vollstrecker:
Seems like we loose 40-50 seconds without ninja. Looks OK when therefor after the failure not warnings from other files pop out.
It's a bit late now, tomorrow I'll add release and debug for mac if noone else feels like it.

PR #269, opened 2021-03-22
Log global state

mrjimenez:


whyman:
Wow, fantastic! I wasnt expecting you do to the work, would have been willing to attempt it myself.

Great work, many thanks @mrjimenez!

mrjimenez:
> Wow, fantastic! I wasnt expecting you do to the work, would have been willing to attempt it myself.

You should have said it, I just did it because things were fresh in my mind. it would be nice to let you see how it works.

In fact, there was one thing that I encounter and did not expect, and this was that before we could enable log before initializing everything. But now we must at least have the UpnpLib struct allocated, so I had to break (for the second time) the UpnpInit2() API to add the log file name. I only caught this because the windows unit test did not produce the log and gave an error in the `type` command. All the other tests had passed.

> Great work, many thanks @mrjimenez!

Thanks!

PR #278, opened 2021-03-29
Work

mrjimenez:


Vollstrecker:
You're sure you want keep this stuff. They don't know about UpnpLib, and rely on pre-generated autoconfig.h
In theory it would be possible to let them use the ones cmake created, but it creates the project-files in the same step also.

whyman:
I think we should start at least talking about dropping autotools, though I imagine Marcelo would not be that keen!

Vollstrecker:
Autohell free world is a dream.

mrjimenez:
> You're sure you want keep this stuff. They don't know about UpnpLib, and rely on pre-generated autoconfig.h
> In theory it would be possible to let them use the ones cmake created, but it creates the project-files in the same step also.

No, I am not sure :laughing: . Just moved it because CMake was building over this directory, and every time I did `rm -rf build` git started to complain about removed files. In fact, I'd rather remove them if no one has any attachments to this legacy.

On a side note, I must admit I have good memories from Microsoft Quick C, but visual C++ is crap, IMHO. I also must confess that, in my opinion, microsoft Visual Basic is the best thing Microsoft has ever done. Unbeatable to the present day to build GUIs. There is nothing comparable in any other platform. The only thing that came close was Borland Delphi/C++ Builder. Unfortunately, the same MS pushed me away from Windows a long time ago...

Back to the topic, lets remove them!

> I think we should start at least talking about dropping autotools, though I imagine Marcelo would not be that keen!

Seriously, I have no problem with that, as long as we move for a better thing. I would like to know what @ffontaine thinks about this, because he is working seriously into Buildroot and that is a project I admire a lot and would not like to create problems for them.

ffontaine:
buildroot has package infrastructures for a lot of build systems: from the venerable autotools (used by 1175 packages), to cmake (used by 288 packages) to the more recent meson (used by 71 packages).

As can be seen in the figures above, autotools is still used by a lot of embedded packages, probably because this is the oldest of them. If you want to switch to cmake, it's fine for me. Howerver, perhaps other build systems should also be considered: for example a lot of open source projects moved (or are moving) from autotools to meson (e.g. gupnp, gstreamer, jsoncpp, ncmpc, systemd, qemu, etc).

Independently of the buildsystem, that you'll use, please kind in mind that pkg-config (i.e. `libupnp.pc` and its `Libs.Private` field for openssl linking) must be kept or I'll have to disable static building with libupnp in buildroot.

Vollstrecker:
> Howerver, perhaps other build systems should also be considered

As cmake is done already and the question is only if two should be kept or only one, this is a bit late.

> please kind in mind that pkg-config (i.e. libupnp.pc and its Libs.Private field for openssl linking) must be kept

This is already in place. Maybe you just give it a try to make sure it works or get's fixed early enough.

@mrjimenez 

As we would need to reconstruct the autoconfig.h generation in this case, as it rebuilds the behaviour of autoheader by parsing the m4-files that would disappear also, this maybe is a good chance to get rid of the generated files also? Except upnpconfig.h for sure.

Vollstrecker:
Oh, and as an addition to meson. This one supports only meson and cmake (which is not guaranteed to work) as subprojects, and they have to be downloaded by hand. So here we would have problems with pthreads, and (as I started because of this) with amule there are many more that have to be installed by hand some way before they can be used. It's a nice idea, but not half as flexible (maybe matured) as cmake.

ffontaine:
> This is already in place. Maybe you just give it a try to make sure it works or get's fixed early enough.

It's broken because `OPENSSL_LIBS` is hardcoded to `-lssl`, `pkg_check_modules` should be called to retrieve  dependencies of openssl such as `-lcrypto`, `-lz` or `-latomic`. I'll send a PR.

Vollstrecker:
${OPENSSL_LIBRARIES} holds the needed information. unfortunately with the full lib path. I would have to read if this would be ok, then the change is easy. But I have to read the pkg-config anyways because of the CMAKE_INSTALL_LIBDIR and relativ path's or such a thing someone mentioned and I wanted t fix 2 weeks ago.

Maybe you know it already?

mrjimenez:
I think he will do it, he just didn't do it because unfortunately we left the CMake configuration out of the 1.14.2 tarball.

Just released 1.14.3.

ffontaine:
Yes, I'm working on it basically, the change is as simple as:

```
	find_package(PkgConfig QUIET)
	pkg_check_modules(PC_OPENSSL QUIET openssl)
	set (OPENSSL_LIBS ${PC_OPENSSL_LIBRARIES})
```

It could be made more complex if we want to support building with openssl on systems where there is no pkg-config support.

Vollstrecker:
I would have gone for regexing over OPENSSL_LIBRARIES to get just the filename without extension and combine that with -l, so we wouldn't need to search for a second time and are independent of pkg-config.

ffontaine:
I'll let you try that path if you want however I don't know if it'll really work. From my experience, cmake buildsystem doesn't really care, handle or perhaps understand the concept of static library linking (i.e. linking with a static library means that you must link with the **full** list of dependencies and in the **correct** order) ...
From my knowledge, the only "standard" way of doing this is through pkg-config (and it should be noted that pkg-config is used with a higher priority than cmake to retrieve dependencies in meson: https://mesonbuild.com/Dependencies.html).

For example, I'm nearly sure that `OPENSSL_LIBRARIES` will never contain `-latomic` (which is needed on some architectures such as sparc or ARM), even `-lz` is probably missing (which is always needed).

That's exactly the kind of issue that I tried to get fixed in cmake upstream projects such as gerbera or wireshark when those packages raise build failures on buildroot autobuilders. But I'm getting a little bit tired to do it over and over so now I'm just disabling static build with cmake packages on buildroot.

But feel free to experiment static linking with buildroot, you'll quickly find the issues. You could even try to send PRs to update https://github.com/Kitware/CMake/blob/master/Modules/FindOpenSSL.cmake.

Vollstrecker:
Jep, it seems that they just know about Threads. Using what we have is the cleaner way, but in this case the working way is better. For fixing their stuff: As they want everything on their own gitlab, I'm not very motivated. Same thing with the inconsistency when preseeding config's to sub-builds when option-names are the same in comparision with the policy to ignore normal vars. Maybe over a mailling-list if I find them.

Have you tried static building upnp on systems without shared libs? I don't expect it to work as of #284, but I'm pretty sure we're getting problems with missing atomic or z there, too as this is not handled atm with the normal targets at all.

ffontaine:
Yes, open_ssl seems to be broken (as reported in #284). I'll also update #283 then we could fix static linking with openssl.

mrjimenez:
On the matter of OpenSSL, please check now after #286 .

Vollstrecker:
I think I need to compare the builds, maybe I miss something somewhere.

Vollstrecker:
@ffontaine: Just had some thoughts about that broken ssl.

- As you say there are some platforms without shared libs, you maybe should move from the config-generation some way up to the general detection and add the needed info to OpenSSL::SSL also. Just a guess, but I'm pretty sure we need it there, too.

- And as you mentioned platforms without pkg-config (I know only of win, but there opensll at all is not that easy), we maybe need some "try with a, try with a,b..."-loop to get the right values.

- And (maybe the biggest issue) I totally forgot about the UPNP-Config.cmake, there isn't even checked for openssl (or if it's needed). I think for this it's the best idea to cache the results and write it to that file like the other config, so the search doesn't need to be done again. Ideally wrapped in some if (NOT TARGET OpenSSL::SSL) to not change something the including project maybe has done before.

PR #283, opened 2021-03-30
CMakeLists.txt: add UPNP_BUILD_{SHARED,STATIC} options

ffontaine:
Building shared libraries are not always possible as some embedded toolchains only support static linking so add options to build shared or static libraries and while at it install headers when building the static libraries

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,6 +82,8 @@ if (WIN32)
 	set (STATIC_POSTFIX s)
 endif()
 
+option (UPNP_BUILD_SHARED "Build shared libraries" ON)
+option (UPNP_BUILD_STATIC "Build static libraries" ON)
 option (BUILD_TESTING "Run Tests after compile" ON)
 
 
--- a/ixml/CMakeLists.txt
+++ b/ixml/CMakeLists.txt
@@ -16,82 +16,88 @@ set (IXML_SOURCES
 	src/nodeList.c
 )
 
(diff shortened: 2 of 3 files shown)


ffontaine:
@Vollstrecker, here is the kind of issue that must be fixed to handle static building with cmake.
It should be noted that this PR is only the first step.
For example, I spotted that the headers are only installed with the shared library and not with the static library. This will result in a broken installation if BUILD_SHARED_LIBS is set to OFF but I didn't fix it as I didn't know if it was intentional .

Vollstrecker:
If you copy line https://github.com/pupnp/pupnp/blob/cd5fe84b09389a04ec1eccf1e18bfa571a524e22/upnp/CMakeLists.txt#L145 under https://github.com/pupnp/pupnp/blob/cd5fe84b09389a04ec1eccf1e18bfa571a524e22/upnp/CMakeLists.txt#L215 the headers get asosiated to both and "should " install always. You could also just move the line as you want the static one always to be build.

I would prefer to use UPNP_BUILD_STATIC and UPNP_BUILD_SHARED. So people have the complete control of what they want to get.

That BUILD_SHARED_LIBS doesn't work is expected. There is one target that explicitly makes a shared lib and one that does the same for a static one. That's needed to get both on one run. BUILD_SHARED_LIBS was meant for situations when you have just one target which doesn't explicitly state what it builds. Then it defaults to static and you tell him that  the default is shared.

PR #291, opened 2021-03-31
Add a LogCallback mechanism

whyman:
This allows client apps to integrate libupnp logging with their own
logging mechanisms


upnp/generator/generator.c
```diff
@@ -175,21 +175,22 @@ static struct s_Member UpnpLib_members[] = {
...
-        INIT_MEMBER(gLogIsStderr, TYPE_INTEGER, int, 0, 0),
-        INIT_MEMBER(gSetLogWasCalled, TYPE_INTEGER, int, 0, 0),
-        INIT_MEMBER(gLogInitWasCalled, TYPE_INTEGER, int, 0, 0),
-        INIT_MEMBER(gLogFileName, TYPE_INTEGER, char *, 0, 0),
+        INIT_MEMBER(LogFp, TYPE_INTEGER, FILE *, "<stdio.h>", 0),
+        INIT_MEMBER(LogIsStderr, TYPE_INTEGER, int, 0, 0),
+        INIT_MEMBER(SetLogWasCalled, TYPE_INTEGER, int, 0, 0),
+        INIT_MEMBER(LogInitWasCalled, TYPE_INTEGER, int, 0, 0),
+        INIT_MEMBER(LogFileName, TYPE_INTEGER, char *, 0, 0),
+        INIT_MEMBER(logCallback, TYPE_BUFFER, LogCallback, "UpnpLog.h", 0),
```
mrjimenez, line 193: Here you can use TYPE_INTEGER, since LogCallback is a pointer.

whyman: Thanks will update it! In the end I handrolled the code, generator wouldnt build as no autoconfig.h present (I was building using CMake)

whyman:
MSVC sucks! C99 is only 22 years old at this point

upnp/generator/generator.c
Vollstrecker: Strange, autoconfig get's generate while parsing the m4-stuff, so it has always to be there. Did you look in the build-dir/?

whyman: Its not built by the buildsystem though, there is just the `compile.sh` script

Vollstrecker: ? Your changing parts of the lib, that should be covered. Beside the gtest I don't know of any compile.sh, and even there they are not needed.

If there are any cases where this is not the case, that needs to be fixed. Just a bit more info is needed.

mrjimenez: There is not much to worry about this specific `compile.sh`, this is currently a developer only thing. Some C source files inside the project are currently generated by a program called `generator.c`. This `compile.sh` script compiles and runs `generator.c` and places the generated files in the correct locations.

In order not to generate unexpected confusion, I have chosen not to put this as part of the build process of the library, although it could be done in that way. Currently it is a strictly manual process that is not supposed to be done every time the program is rebuilt. Also, there are plans to remove this once we have everything "C-plus-plus-ed".

The part of the API that this script generates has been stable since `1.8.x`. Only recently it has suffered some addition because of the new UpnpLib object we are currently messing with.

Vollstrecker: OK. But it sounds like it's a bit messy if you have to choose to build, or to have use that. It would be easy to make an option like USE_GENERATOR (DEV ONLY) and then call the generator and use this, or just do what is done now.

Vollstrecker: > I fully agree with you. It is messy. But do notice that CMake is around the block for as much time as you are

Yes I know. It's the same I was told about the debian-dir in amule back in 2004.

> O course, if it is just too easy...

Running commands is just a command, and placing the sources in vars is just copy and past. Mayb 10 Minutes to create the generator. And yes, running on Unix is enough as we see here. The question is just: "Can the generator run only on Unix?" If yes, you'll get it only there. If yes, not going to touch it. If no, I'm not going to touch it.

And yes, prio for that is low.

mrjimenez: > > I fully agree with you. It is messy. But do notice that CMake is around the block for as much time as you are
> 
> Yes I know. It's the same I was told about the debian-dir in amule back in 2004.

I remember. :laughing: 

> Running commands is just a command, and placing the sources in vars is just copy and past. Mayb 10 Minutes to create the generator. And yes, running on Unix is enough as we see here. The question is just: "Can the generator run only on Unix?" If yes, you'll get it only there. If yes, not going to touch it. If no, I'm not going to touch it.

As far as I can be sure (because I tested), works only in linux.

> And yes, prio for that is low.

Sure. :+1:

PR #304, opened 2021-04-02
Internal unit tests

mrjimenez:
The idea is to gather all internal tests in one place.

Unit test from an external point of view does not apply because the
tested functions are not necessarily in the API.


Vollstrecker:
2 thoughts on this:

- You think it's a good idea to have the test-code in the lib itself? I would prefer to define the struct etc. in the test-code, or is is going to be used somewhere productive?

- You sorted the test in the CMakeList between url and upnpstring. I prefer alphabetical ordering, or just appending, but that order?

mrjimenez:
> * You think it's a good idea to have the test-code in the lib itself? I would prefer to define the struct etc. in the test-code, or is is going to be used somewhere productive?

These are internal tests, they cannot be performed from outside in a test code linking to the library because the functionality is not exposed in the API.

What I believe that should be done is to include a ~~configure directive~~ CMake option :laughing: , to enable or disable the unit test code, since it is not used during normal use.

> * You sorted the test in the CMakeList between url and upnpstring. I prefer alphabetical ordering, or just appending, but that order?

Please do sort it in the way you think is most appropriate or more to your taste, I have no preferences about it.

Regards,
Marcelo.

PR #309, opened 2021-04-06
GTests02

ingo-h:
Skip tests if running as github workflow actions: it is possible to check an environment variable on the github workflow runner and skip the test if it exists.


Vollstrecker:
Moin,

first the important. With *.cmake in ignores, you block the whole contents of cmake/ dir. I guess your running config and build in the source-dir. This works, but cmake itself was explcitely designed that this is not needed. For example my project structure:

```
upunp
  |----pupnp (the source)
  |----build-release
  |----build-debug
  |----build-release-openssl
```
So I configured them one with cmake ../pupnp and that's all. I'm too lazy for cmake --build ., so I just type make as it's shorter on this platform. This way you get code changes after pull or branch-change, plus cmake is automatically rerun if something changes there. Plus I don't need to ignore files on commit.

Now to the tests itself. Theses changes indeed make it easier to run the tests at home. My idea (and Marcelos if I didn't missunderstand) was to have them in the actions to see if something changes. With this solution you'll never see any change, without changing the code of the tests. If this is intended, I would at least expect any type of TODO-comment, so the places for the change can easily be found if needed.

ingo-h:
> first the important. With *.cmake in ignores, you block the whole contents of cmake/ dir.

What do you suggest? Shall I remove

    *.cmake
    CMakeFiles/

from the local `.gitignore` file?

>My idea (and Marcelos if I didn't missunderstand) was to have them in the actions to see if something changes. With this solution you'll never see any change, without changing the code of the tests. If this is intended,

No, it isn't. The tests are self documented similar to the library with doxygen. Of course you have to put the info into the tests and then the comments are shown with the test result on the console. But there is a conflict with the Github Workflow. Either the test succeds, then the console output is suppressed, or it fails and you can see the console output.

>I would at least expect any type of TODO-comment, so the places for the change can easily be found if needed.

These TODO-comments are available. Normally there is nothing suppressed if local executed. All failures are shown so you can fix them. If you want to see all Github Action Skips and its places in the test just execute it with, e.g.:

    GITHUB_ACTIONS="true" ./test_template-static

as documented in the local README.md.

Vollstrecker:
> What do you suggest? Shall I remove

At least *.cmake. CMakeFiles would be OK, they just shouldn't occur if you use cmake how it was designed.

> >With this solution you'll never see any change, without changing the code of the tests. If this is intended,
>No, it isn't. The tests are self documented similar to the library with doxygen.

I wasn't talking about the documentation. I meant that even if this is fixed, someone would have to remove the env-check to get them running in Actions. Given, if you would split them into a separate file, this would mean to move them back as this is fixed, too. But this is because of googletest is designed to have a bunch of tests in one file. I still prefer one file per test, so we could easily just remove the WILL_FAIL flag without touching any code.

>Of course you have to put the info into the tests and then the comments are shown with the test result on the console. But there is a conflict with the Github Workflow. Either the test succeds, then the console output is suppressed, or it fails and you can see the console output.

Not really. You don't get the output because this of --output-on-failure only of there is a failure. If you run ctest -V, your get all the output. I don't think this is very useful in actions, because this is getting harder readable with each test added. If we could mark the failing tests, it wouldn't be a problem to just change prefix the failing ones and then run the succeeding, and afterwards the failing. Because of the marking that they fail this will still get a green status, but the failing ones would get output to diagnose.

This is not just useful for the (unlikely) event that someone fixes that as a sideeffect without recognizing this (I don't run the tests, as my test is the successful compile), everyone that run the test now has to remember at least how many tests fail to see if something change. The big advantage is to see if they work the same way on every run on every platform. You had in the first version an uninitialized pointer. I guess this worked for you and was only found as they began to randomly fail and succeed.

> That is what should come as action output if it wouldn't suppressed.

That is what comes if we could run them with -V. But instead of telling everyone that there is a problem with gmock, I would like to see more specific stuff. The other tests tell me what the expected and what they got, so there is a chance to look for it. Without that skip, maybe someelse get's an idea.

Oh, and I still don't get the meaning of the template. Iirc this is an example that does nothing, so anyone who would want to start with a test would need to copy this to a new source-file. I hardly convinced that anyone doing something with gtest should be able to write that little boilerplate before he's actually thinking about writing a real test. It doesn't harm, and I don't care if it is in there or not. I just don't understand the need of it.

mrjimenez:
Hi ,

> Now to the tests itself. Theses changes indeed make it easier to run the tests at home. My idea (and Marcelos if I didn't missunderstand) was to have them in the actions to see if something changes. 

Correct.

@Vollstrecker , is this changeset now ok to commit? If you're still missing something, post a (numbered) list so that we can check. Otherwise give me the green flag and I commit.

mrjimenez:
Hum, it seems like @Vollstrecker last commit interfered with this one :disappointed:

Vollstrecker:
The first commit interfered. Now it doesn't conflict, but in gtests/CMakeLists.txt 3/4 of the line are indented one level too much. That's atm the only thing I see.

This changeset makes it better. Maybe I'll take a cloder look at the splitting later, but first I have 5 or 6 other things on my mind. I'll stick to one piece a day for now, as it's amule week.

So from my side it's green.

mrjimenez:
Ok, just tested it but I am getting confused. The tests in Github actions give no errors. But when I run ctest here, I get:

```bash
...
gtest/test_upnpapi.cpp:295: Failure
Value of: captFdObj.print(std::cerr)
  Actual: true
Expected: false
Output to stderr is true. There should not be any output to stderr
[  FAILED  ] UpnpApiIPv4TestSuite.initialize_default_UpnpInit2 (52 ms)
[----------] 1 test from UpnpApiIPv4TestSuite (52 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (52 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] UpnpApiIPv4TestSuite.initialize_default_UpnpInit2

 1 FAILED TEST

      Start 37: test-upnp-UpnpApiTestSuite.get_handle_info-static
37/38 Test #37: test-upnp-UpnpApiTestSuite.get_handle_info-static ...................................   Passed    0.01 sec
      Start 38: test-upnp-UpnpApiTestSuite.get_error_message-static
38/38 Test #38: test-upnp-UpnpApiTestSuite.get_error_message-static .................................   Passed    0.01 sec

89% tests passed, 4 tests failed out of 38

Total Test time (real) =   0.53 sec

The following tests FAILED:
         30 - test-upnp-UpnpApiIPv4TestSuite.UpnpGetIfInfo_called_with_unknown_interface (Failed)
         31 - test-upnp-UpnpApiIPv4TestSuite.initialize_default_UpnpInit2 (Failed)
         35 - test-upnp-UpnpApiIPv4TestSuite.UpnpGetIfInfo_called_with_unknown_interface-static (Failed)
         36 - test-upnp-UpnpApiIPv4TestSuite.initialize_default_UpnpInit2-static (Failed)
Errors while running CTest
```

What am I doing wrong?

mrjimenez:
Another problem. There are 38 tests being run on unix systems, but only 28 on windows builds. The number should be equal.

If a test does not run in a system, it should be considered as failing. The separation cannot be by operating system.

ingo-h:
I haven't used CTest. Please look at the README.md.

Vollstrecker:
Look at the code. https://github.com/pupnp/pupnp/blob/f50a5609d3228b2a8c8169c42b37ad99e35d07c7/gtest/test_upnpapi.cpp#L156. This means basically "if on github return true". That's what I was talking about in the comment you cited. That's why I think separating them into a different file is the best way. That way they can be run and the failing of the test can be judged as success, so you can see that it's still broken, but you also see instantly when this changes.

For the separation by system: They don't even build on win (at least earlier versions), so either this has not changed, so the separation is ok atm, or this has changed and he just forgot to remove the if(NOT WIN32) which does exactly that separation.

PR #317, opened 2021-04-08
Be more specific about what to install where and when.

Vollstrecker:
Doesn't change that much, but when using cmake (the executable) to install, it's now possible to just get the libs or the dev parts. Could be usefull for distro's where the packages arwe split.


whyman:
Can we go with `UPNP_` to match the targets?

PR #348, opened 2021-12-12
CMake: Add LFS definitions when needed

whyman:
Previously we worked out if we needed them, but then they werent set


whyman:
No idea whats happening in the CI, unrelated it seems

Vollstrecker:
Rough guess, they changed master to something more political correct. Btw.
I'm no friend of add_compile_definitions, I would suggest setting the corresponding var for LFS and add a conditional gen-exp where needed. Just to be sure the flag doesn't propagte to other targets when we are built as a subbuilt.
I can't give a clearer explanation atm. but I hope I get the time to adress the various stuff after the 20th if my words are to weird to follow (wouldn't be surprised).

Vollstrecker:
The CI problem was solved last week. I created a pr that makes the flags part of ixml interface, so everything should propagate up to the last binary.

PR #349, opened 2022-01-06
Use the values for LFS that where detected.

### LINKED ISSUE #345: libupnp built with cmake is not functional
Hi,

There has been a lengthy discussion/investigation in https://github.com/gerbera/gerbera/issues/2000. But it has boiled down to the fact that when I configure libupnp with cmake, then gerbera is not functional. Whenever I configure libupnp with autoconf, then gerbera works fine.

I'm using Arch Linux ARM on armv7.

libupnp versions that I've tried are 1.14.12 and 1.14.7.

If there is additional information that you need to help solve this problem, I'm ready to provide it.

I want to add that tests (make check) pass no matter how I configure - with cmake or autoconf.

Vollstrecker:
Fixes #345
Replaces #348


whyman:
Looks sensible to me

Vollstrecker:
Dumb question: If we have them in the compiler-flags, is the definition still needed in upnp/inc/upnpconfig.h and upnp/sample/common/config_sample.h?

mrjimenez:
I think ubuntu-16.04 is no longer supported, is that so? If true, we should remove it.

Vollstrecker:
That's true and on my list for after the merge. We want to wait till @kyak confirmed that it works, or just assume it?

mrjimenez:
Since this is old stuff, lets commit it and wait for further complaints.

kyak:
You can as well assume. I'll be able to confirm in a couple of days

mrjimenez:
> I think ubuntu-16.04 is no longer supported, is that so? If true, we should remove it.

Oops, just realized I did it in september.

mrjimenez:
Do you mind porting it to 1.14.x? I am affraid to screw things on CMake, its been a while.

kyak:
> Do you mind porting it to 1.14.x? I am affraid to screw things on CMake, its been a while.

Please do so, otherwise I won't be able to test. Gerbera currently doesn't support libupnp 1.16 afaik.

Vollstrecker:
> > I think ubuntu-16.04 is no longer supported, is that so? If true, we should remove it.
> 
> Oops, just realized I did it in september.

Is the commit-date shown here the one when you fixed it without pushing, or why did it run here earlier?

mrjimenez:
Sorry, I did not understand your question. The commits fixing the ubuntu 16-04 issue are:

master: 135527d6ed990407c2750e81a6ef0feb1b105120
branch-1.14.x: f9c8f5d4d817f58dfc81b05a575c70ce2afe3bdc

Vollstrecker:
The question was, if the dates in the commits refer to the date when you committed the changes, and then you pushed a little (3 month) later as They shouldn't run after the changes, but I'm sure I've seen them running when I committed this stuff here.

mrjimenez:
I don't remember pushing it three months later, I usually push ASAP, but since I don't remember, it might have happened, so I am sorry for that. The last three months of last year have been complicated for me.

The dates in the commits are part of the commit, they enter the SHA1 hash, so they are the dates when the commits were performed, they are not the date of the push.

I don't know how or where to look for the push date, if you can tell me I will sincerely appreciate it. :smiley: 

Regards,
Marcelo.

Vollstrecker:
I don't think this is worth the time to investigate further. It's possible that I caused it, my branch is based on 8c73f9c9ebd8 which is something older than the changes, so propably this caused it.

PR #354, opened 2022-01-08
Disable Blocking TCP, Enable reuseaddr by default

kyak:
and remove the crazy POST write "functionality"

This is a backport of 4265d32eb4169428192c242cfadea8e3d5aa92ee to 1.14.x branch


Vollstrecker:
I'm not sure if this is something that should be changed in a stable release-cycle.

kyak:
That would be very convenient to change default values of these two options in 1.14.x branch. This way libupnp will be built with the recommended (by gerbera) options by default.

If this is not possible, I will file a bug report at Arch Linux, because they don't currently specify these options: https://github.com/archlinux/svntogit-packages/blob/packages/libupnp/trunk/PKGBUILD

I had assumed that other distributions could also benefit from these changes, but it looks like Arch and Gentoo are the only distributions who package 1.14 branch, and Gentoo already specifies these flags (although they chose to use blocking-tcp-connections for whatever reason).

Vollstrecker:
Maybe it would be a better approach to change this from build to runtime option. An additional parameter for init would be enough for this and the configoption would just influence the default if no option is given. That way the api would stay almost stable and everyone is independent of what the distro thinks is better.

kyak:
I think that's a good idea to change this to runtime option. But this won't get to 1.14 branch, since this is API change.

Anyway, I've filed the bug report at Arch, for your reference: https://bugs.archlinux.org/task/73298

Vollstrecker:
I wouldn't say upstream isn't going to change that, as I (if at all) have only one vote of at least 3. And what the code itself or strategic decissions regards my opinion is more a rough guess. Even if he denies that, the last instance is Marcello and Ian is also a better advisor in such cases than I am.

mrjimenez:
I tend to agree with @Vollstrecker that maybe this is not something to do in a stable branch. But this is not lost work. I have some doubts about if we did the right thing with the library handle thing, but I have not had the necessary time to think about it. Maybe we should open a discussion about it.

PR #355, opened 2022-01-08
Dont call miniserver callbacks for failed hosts

whyman:
When we implemented validation in Gerbera[0], I noticed that even if our
callback returns UPNP_E_BAD_HTTPMSG the request still get processed and
not dropped as expected.

Practically that means all hosts are valid if you have a validation
callback set :(

0. https://github.com/gerbera/gerbera/pull/2365


Vollstrecker:
Sorry for the failed tests, I changed a bit to not build what is not needed and didn't check for a second build.

PR #358, opened 2022-01-12
CI: Add msys2/mingw builds

Biswa96:


Biswa96:
This is just a basic build workflow in msys2. This can be extended with more options.

mrjimenez:
Feel free to improve it.

Thanks!

mrjimenez:
@Vollstrecker , @whyman , @Biswa96 ,

Anyone can think of a reason not to copy ccpp.yml from master to branch-1.14.x? Because I am really tempted to do so...

Vollstrecker:
I guess the script-file called in Lines 52 and 81 will give you an error as it is just not there, but without the lines it's just configuring, building and testing, so this is something that has to work.

PR #360, opened 2022-01-13
Use a workflow closer to the one used in master

mrjimenez:


Biswa96:
The mingw builds need patches from my previous pull request.

mrjimenez:
Can you add them?

Biswa96:
Did you mean to create a pull request with that same change but for 1.14 branch?

mrjimenez:
Sorry, I somehow messed it up. I though I was using a copy of the file after your changes. I'll close this one and do another with your changes.

mrjimenez:
In fact, I did not messed it up :smile: I think I don't know what are the changes you refer to.

mrjimenez:
The file ccpp.yml in this branch is the same as the file in the master branch, except for the two lines @Vollstrecker said that should be commented.

Biswa96:
I said that the CMake changes are required for the mingw CI build.

Vollstrecker:
I think he means 2226adc72079e7a54c59171f9378 but I also think this one is backported.

mrjimenez:
I might have messed up in the backport, anyone care to check?

Vollstrecker:
It seems I got it, this one was a bit harder as cmake complained about line 19 where the wrong value was used and not set. But if I see this correct, the teststep itself is missing in mingw.

Biswa96:
I did not added tests in CI because many tests failed. Three options I had:
1. disable tests entirely.
2. enable tests but make CI happy without the failure.
3. enable tests and CI will be red if any test failed.

I selected the first one. Hence, I commented the pr as "basic build workflow".

Vollstrecker:
k, are you going to fix or at least report the fails? As for now it seems that you're the only one seeing them.

Biswa96:
I can try.

Vollstrecker:
From what I recall, it's not clear everytime what lib is missing. Does mingw set an rpath on the binaries? If not, something like the env-settings in the before the tests could be needed there also.

Biswa96:
In msys2/mingw environment, the PATH variable looks like this:

```
$ echo $PATH
/mingw64/bin:/mingw64/bin/site_perl/5.32.1:/mingw64/bin/vendor_perl:/mingw64/bin/core_perl:/usr/local/bin:/usr/bin:/bin:/c/Windows/System32:/c/Windows:/c/Windows/System32/Wbem:/c/Windows/System32/WindowsPowerShell/v1.0/
```

/mingw64/bin contains all the .exe and .dll files.

Vollstrecker:
Sure, but the test-progs also need libupnp which is not in these paths, On Linux systems the somewhere in the bin is a path noted where the libs have been on linking time, the rpath. With that info it can run without installing the lib previously. On Windows this isn't given, so we had to add the path's of the libs to PATH before the tests.

I hardly guess in mingw it's the same, as you say the static ones run.

PR #363, opened 2022-01-14
Drop ithread header

whyman:
At this point its a very thin wrapper around pthread anyway.

Tests pass on linux/macos.


mrjimenez:
I think you will need to create a symlink like:

```bash
ln -s  upnp/src/inc/sleep.h upnp/sample/common/sleep.h
```

Vollstrecker:
Or adjust the #include and include-dirs, that way it would work on plattforms without symlinks also.

mrjimenez:
I don't know. The file sleep.h is a tool from libupnp, it is not distributed. To use it, one must copy it.

What happens in windows when you checkout a symlink?

Vollstrecker:
> I don't know. The file sleep.h is a tool from libupnp, it is not distributed. To use it, one must copy it.

So it shouldn't be used in the samples either, I guess. Otherwise it would be a bad sample as it's not transferable. But getting the includes to point to the original file shouldn't be that hard.

> What happens in windows when you checkout a symlink?

Never tried. Unfortunately I'm in the middle (optimistic me, in reality around 20%) of doing my taxes, so I don't think I can do much on this for at least the next week.

Vollstrecker:
Nice that you added the suppression. I wanted to try the "fixes" the compiler suggests, now I have to even remember that there is something like that.

But the CI get's better readable with this in there.

PR #364, opened 2022-01-14
CI: Add tests in msys2/mingw builds

Biswa96:


Biswa96:
The statically linked test binaries runs successfully but the dynamically linked ones do not. The `0xc0000135` error code in test means the program can not find a DLL. I have to copy libstdc++-6.dll, libgcc_s_seh-1.dll and libwinpthread-1.dll files from `$MSYSTEM/bin` to build/upnp/test folder for those tests.

Vollstrecker:
gtest shouldn't be needed, I guess I have to get mingw to work here to check what's going on.

Biswa96:
According to Process monitor, the dynamically linked tests can find libixml.dll and libupnp.dll from build directory. But it can not find those previously mentioned dlls. Here is the a portion of the output from Process Monitor:

```
test-upnp-init.exe	5852	QueryOpen	F:\pupnp\bin\upnp\test\libgcc_s_seh-1.dll	NAME NOT FOUND	
test-upnp-init.exe	5852	CreateFile	C:\Windows\System32\libgcc_s_seh-1.dll	NAME NOT FOUND	Desired Access: Read Attributes, Disposition: Open, Options: Open Reparse Point, Attributes: n/a, ShareMode: Read, Write, Delete, AllocationSize: n/a
test-upnp-init.exe	5852	CreateFile	C:\Windows\System\libgcc_s_seh-1.dll	NAME NOT FOUND	Desired Access: Read Attributes, Disposition: Open, Options: Open Reparse Point, Attributes: n/a, ShareMode: Read, Write, Delete, AllocationSize: n/a
test-upnp-init.exe	5852	CreateFile	C:\Windows\libgcc_s_seh-1.dll	NAME NOT FOUND	Desired Access: Read Attributes, Disposition: Open, Options: Open Reparse Point, Attributes: n/a, ShareMode: Read, Write, Delete, AllocationSize: n/a
test-upnp-init.exe	5852	QueryOpen	F:\pupnp\bin\upnp\test\libgcc_s_seh-1.dll	NAME NOT FOUND	
test-upnp-init.exe	5852	QueryOpen	F:\pupnp\bin\ixml\libgcc_s_seh-1.dll	NAME NOT FOUND	
test-upnp-init.exe	5852	QueryOpen	F:\pupnp\bin\upnp\libgcc_s_seh-1.dll	NAME NOT FOUND	
test-upnp-init.exe	5852	QueryOpen	F:\pupnp\bin\upnp\test\%PATH%\libgcc_s_seh-1.dll	PATH NOT FOUND	
```

Vollstrecker:
This looks like the exakt opposite of what I expected. Could you post the upnp/test/CTestTestfile.cmake, maybe this gives some insights.

Biswa96:
This one? [CTestTestfile.cmake.txt](https://github.com/pupnp/pupnp/files/7873087/CTestTestfile.cmake.txt)

Vollstrecker:
Exactly, you see the PATH=.... in 8? I guess %PATH% is not what mingw expects, you can edit this file directly, if this works, you could mask [this line](https://github.com/Biswa96/pupnp/blob/839ec0fc26cb8aaca1ad2416e29d770d2ba61d68/cmake/test-functions.cmake#L122) with if (MSVC) and add an else with the value that works for you.

Vollstrecker:
Maybe mingw is smart enough to find it's stuff if you remove that whole path setting thing, would make the fix easier, too.

Biswa96:
Thank you for the hint. I have limited that environment searching function for MSVC only and added the recently built DLLs path with ctest command. Now tests passed in mingw.

whyman:
Does 
`D:/a/pupnp/pupnp/upnp/src/threadutil/ThreadPool.c:172:9:` warning: implicit declaration of function 'gettimeofday'; did you mean 'mingw_gettimeofday'? [-Wimplicit-function-declaration]
  172 |         gettimeofday(&now, NULL);
      |         ^~~~~~~~~~~~
      |         mingw_gettimeofday`

Need to be addressed?

Biswa96:
That looks like a different topic to me. I am thinking to address that in different pull request with those string and I/O redefinition warnings. Just some ifdefs.

PR #370, opened 2022-01-22
14.x CMake: Suppress MSVC warnings

whyman:


Vollstrecker:
Maybe this would be better set private and added to the other lib also. If someone chainbuilds his stuff with the lib, this would propagate to their stuff and maybe he doesn't want this.

whyman:
Perhaps, this is just a backport from master.

Vollstrecker:
I remember, so we should either change that on master and backport then, or merge and fix this and forwardport it.

Vollstrecker (commented):
I don't get it., empty line on 24, _CRT* still public, and I can't remember script_support was removed.

Am I right that your'e still trying to get everything in order?

PR #373, opened 2022-01-28
CI: Comment reference to nonexistant script

whyman:


Vollstrecker:
Wouldn't it be easier to copy the file from master? If current 1.14 will get master the other stuff will be ported there also, this would be a first step.

PR #374, opened 2022-02-03
Cmake backports

Vollstrecker:
Now almost everything should be on the same level as master.
GTest-support is deactivated for now, as these checks seem to didn't make it to this branch.


whyman:
LGTM!

whyman (approved):

kyak:
I seem to have problem running `make install` with this commit:

```
CMake Error at upnp/cmake_install.cmake:99 (file):
  file INSTALL cannot find
  "/home/user/build/libupnp/src/libupnp-1.14.12/upnp/inc/UpnpHttpHeaders.h":
  No such file or directory.
Call Stack (most recent call first):
  cmake_install.cmake:85 (include)
```

kyak:
> Right, this file is not existing - Fixed.

This one is missing as well: `upnp/inc/UpnpLog.h`

PR #375, opened 2022-02-05
Readd g tests

Vollstrecker:
I added the gtests from master here, 2 out of 3 are not compilable due to lack of UpnpLib (obviously)
The remaining test is basically a tech-demo which checks nothing.

Is there stuff tested that with the handle present and not covered yet? If not, I would prefer to drop the whole gtest instead of having just dummies in there. If yes, someone that understand the lib (not me) could adapt them.

If there's (atm) no point in having gtest present, the requirement C++1 could be dropped also, which should speed up config-process a (tiny) bit.


whyman:
My 0.02 is that there should be some kind of "real" test framework available to use.

Vollstrecker:
Does that mean the empty test should go on, or that we have unittests and don't need a second one?

mrjimenez:
The C++ stuff should go into the next master as soon as possible, so to spare you some work, I don't mind leaving it there.
As for the test framework, I agree, a real test framework is better than spraying tests around the code. So, even if totally unnecessary in 1.14.x, it is also sparing work from the future, what do you think @Vollstrecker ?

Vollstrecker:
I'm neutral in this point. atm. there is one dummy-check, but at least we see it could work. I'm unsure about the C++11, if it is needed for gtest, or if it was for the lib-handler. It would produce an error if the compiler is not capable of this feature, but as it is 11, I don't think there can occur real problem. The time for the additional check is somewhere under a second.

If the checks can be adjusted to work, or this will be useful in the future, I'm fine with it. I just provide the base, you guys do the magic.

mrjimenez:
So, I assume the pull request is ok, right? Can we commit it?

Vollstrecker:
From my side it's complete. When it's merged I can continue with shortening the lists, this conflicts with the changes here.

PR #376, opened 2022-02-09
Visibility changes

Vollstrecker:
As discussed in #372 here are some changes for review


Vollstrecker:
Looks like it's done. Although I prefer closing brackets on a line of their own when over multiple lines.

whyman:
Does autotools also need updating?

whyman:
Does autotools also need updating?

Vollstrecker:
I don't think so. I hardly guess the LIBUPNP_EXPORTS and UPNP_STATIC_LIB comes from them, or was set there already, as otherwise noone ever would have used them. And libtool works different, they intercept the linker calls, grep for the regex the functions to export must match and give a map to the linker to tell him what to export.

atm. I'm searching through their scripts where the version arithmetic's are done, if I can replicate this in the configure script, at least libtool can be dropped from the deps.

Vollstrecker:
k, just for update: libtool is not removeable. I can let automake create the libs without lt, but then it doesn't allow to install it in lib/. It could work if I bypass automake also, but this would result in some kind of self-written configure to make it really useable. In this case all autotools would be gone, but an additional script has to be maintained.

So it's cmake and autotools or cmake-only, but this is nothing to think about in a stable-branch.

PR #383, opened 2022-02-17
Warnings 2

Vollstrecker:
This should fix the remaining warnings.
This is still work in progress, but as I'm sometimes unsure (especially if the last one works as expected) I create this pr now to gather feedback.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -40,6 +40,7 @@ jobs:
           -DBUILD_TESTING=ON
           -DDOWNLOAD_AND_BUILD_DEPS=ON
           -DUPNP_ENABLE_OPEN_SSL=ON
+          -DCMAKE_C_FLAGS="-Wall -Wpedantic"
           -DCMAKE_INSTALL_PREFIX=${{ GITHUB.WORKSPACE }}/test-install
           ${{ env.cmake_extra_flags }}
       - name: Build
--- a/gtest/test_template.cpp
+++ b/gtest/test_template.cpp
@@ -27,7 +27,13 @@ TEST(EmptyTestSuite, empty_gtest)
 	//GTEST_SKIP_("to show this feature");
 
 	// SKIP on Github Actions
-	char* github_action = std::getenv("GITHUB_ACTIONS");
+#ifdef _WIN32
+	char* github_action;
+	size_t len;
+	_dupenv_s(&github_action, &len, "GITHUB_ACTIONS");
+#else
+	char *github_action = std::getenv("GITHUB_ACTIONS");
+#endif
 	if(github_action) { GTEST_SKIP()
 		<< "  to show this feature";
 	}
--- a/ixml/inc/posix_overwrites.h
+++ b/ixml/inc/posix_overwrites.h
@@ -12,6 +12,7 @@
 	#define strnicmp _strnicmp
 
 	/* Secure versions of functions */
+	#define strcat(arg1, arg2) strcat_s(arg1, sizeof(arg1), arg2)
 	#define strcpy(arg1, arg2) strcpy_s(arg1, _countof(arg1), arg2)
 	#define strncpy(arg1, arg2, arg3) strncpy_s(arg1, arg3, arg2, arg3)
 	#define sprintf(arg1, ...) sprintf_s(arg1, sizeof(arg1), __VA_ARGS__)
--- a/ixml/inc/posix_overwrites.h
+++ b/ixml/inc/posix_overwrites.h
@@ -12,6 +12,7 @@
 	#define strnicmp _strnicmp
 
 	/* Secure versions of functions */
+	#define strcpy(arg1, arg2) strcpy_s(arg1, _countof(arg1), arg2)
 	#define strncpy(arg1, arg2, arg3) strncpy_s(arg1, arg3, arg2, arg3)
 	#define sprintf(arg1, ...) sprintf_s(arg1, sizeof(arg1), __VA_ARGS__)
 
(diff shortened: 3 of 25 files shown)


mrjimenez (approved):
Seems ok for me.

Vollstrecker:
So now are just 2 things left.

1) The GetVersionExA call in httpreadwrite.c
For this deprecated function there seems to be no replacement that reports the version number. If I see this correct, with a manifest we could continue to use it, but is it worth the effort for just printing the windows version? I don't get the meaning of unspecified server, but I guess unspecified vs. UPNP_VERSION_STRING should do the trick as good as if the windows version would be prepended.

2) The localtime call in upnpdebug
timeinfo is declared with struct *tm, as tm is declared in a included header of the header that declares localtime (and localtime_s as replacement), so usually the struct could be omitted to get the real struct instead of an empty placeholder. This doesn't work, without struct I get undeclared variable error. If I keep the struct in there, I get a warning about uninitialized usage of this var (in the function that should initialize it), if I try to initialize the struct on declaration, or later member-by-member this warning just goes up to this lines. Anyways, even when copying the examples MS gives in the docs, init-test segfaults exactly on this function. So I'm stuck at this without even understanding what happens there.

mrjimenez:
Hi @Vollstrecker 

Nice work!

> So now are just 2 things left.
> 
> 1. The GetVersionExA call in httpreadwrite.c
>    For this deprecated function there seems to be no replacement that reports the version number. If I see this correct, with a manifest we could continue to use it, but is it worth the effort for just printing the windows version? I don't get the meaning of unspecified server, but I guess unspecified vs. UPNP_VERSION_STRING should do the trick as good as if the windows version would be prepended.

Indeed, a complete mess they did here. IMHO, this could just return "Windows". Until someone comes up with a better way to identify the OS without having to write 50 lines of code in useless `if` statements.

> 2. The localtime call in upnpdebug
>    timeinfo is declared with struct *tm, as tm is declared in a included header of the header that declares localtime (and localtime_s as replacement), so usually the struct could be omitted to get the real struct instead of an empty placeholder. This doesn't work, without struct I get undeclared variable error. If I keep the struct in there, I get a warning about uninitialized usage of this var (in the function that should initialize it), if I try to initialize the struct on declaration, or later member-by-member this warning just goes up to this lines. Anyways, even when copying the examples MS gives in the docs, init-test segfaults exactly on this function. So I'm stuck at this without even understanding what happens there.

That one I didn't understand. The code is supposedly independent of the Windows platform, localtime() is in the posix manual. So, it should just work the way it is. What is the warning you get?

Vollstrecker:
> That one I didn't understand. The code is supposedly independent of the Windows platform, localtime() is in the posix manual. So, it should just work the way it is. What is the warning you get?

MS and posix don't agree always. I guess the posix-functions are just fixed as drop in replacement, and on Win they made new, saver, functions.

As you can see on the other commits they usually change *out func(*in...) to something like error_t = func_s (*out, *in...) the docs say that localtime and localtime_s make a tm out of a time_t, so no problems there. Both functions are defined in the same header, That's Why I'm so confused that there is no way to make it work.

I changed the version-string, Universal...on windows so now only the 2 warnings for localtime are left. I could deactivate them with a define, but I would prefer to fix this, too. If anyone finds out how, so to keep it in mind I'll leave that as is, so anyone can see it.

mrjimenez:
Ok, no hurries. I'll be out for a week without a computer, so I will be quiet for a while.

Regards,
Marcelo.

Vollstrecker:
It's been a bit uncomfortable, but now all warnings are gone. I'm unsure, as it's now the time to increase the warning level and if everything then is fixed set warnings as errors, but this would mean to get on Win ~320 from pthreads if I don't want to set this in CMakeLists.txt (I don't want this, as internal testing stuff shouldn't go in there). Even on Linux this would mean a masssive amount of warnings with -Wall -Wpedantic.

So I would say this part here is done and can be merged.

whyman:
Possibly enabling `-Wall -Wpedantic` for development builds would be a good compromise?

Vollstrecker:
I think this would be the only possible compromise. On Win we get 241 warnings with /W4 (instead of W3 now), 2/3 are from pthreads. With /Wall this goes up to ~6900 including the MS-headers itself (which I don't even think about to fix). I guess most of the warnings (36 atm.) on the other platforms are in common parts and will be fixed for win on the same change.

Vollstrecker:
k, after realizing that msys uses the ms compiler, and then realizing that pt4w is needed there also, the higher warnings are only on Linux active. Ubuntu 18 is not really picky, but the 20 version gives some hints. If someone fixed them someday, I would vote for setting -Werror also, so we don't get this mess again.

Vollstrecker:
So @whyman , have you checked the changes don't break anything in gerbera? Not that any return has changed in a way it can't handle.

whyman:
There are no changes in the public interfaces, so it should be fine.

Vollstrecker:
I don't know how public the versionstring is that is changed in #3fdabf2f01bab6a. I could imagine this is used somewhere. So if not, this stuff here can get merged.

Vollstrecker:
And, as there are many changes that affect memory-management (more or less indirect), a run with a prog that uses it could bring problems to daylight that are not covered by our tests.

mrjimenez:
> There are no changes in the public interfaces, so it should be fine.

There is a pending review, should I assume all is fine on your side and commit?

Regards,
Marcelo.

Vollstrecker:
From my side I'm done. The review is from Ian, if he wants to give it a try first depends on him, if not I would take his comments as approval.

neheb:
@Vollstrecker any reason why the inet_pton change is limited to Windows?

Vollstrecker:
Rough guess: The other way worked without warnings on the other systems so I searched a way of doing it on Win and just didn't check if that way would be useable on the others also.

PRs are always welcome if they take complexity from the code.

PR #390, opened 2022-04-28
Fixes double free caused by closing a file pointer twice

frobones:
Using valgrind (https://valgrind.org/) I tracked down the issue of a double free to the following line in `webserver.c`: https://github.com/pupnp/pupnp/blob/master/upnp/src/genlib/net/http/webserver.c#L554

It appears that `fp` is closed and then is closed again at the `exit_function` label.

Here is the commit that introduced the double close: https://github.com/pupnp/pupnp/commit/0b1a6c59be3babc3ecaf81debdfc159759872215#diff-0b6cfade43473f4ba934e72d8cf4618e511b14e5aefdc60b5ca7a2a6ee18425f

Setting `fp` to `NULL` after it is closed indicates that `fp` is no longer needed in this scope and prevents the second `fclose` from happening.


mrjimenez:
Hi Stephen,

Good job, just merged your patch.

Please, do nou use master, we will continue development from 1.14.x. I know, this is my fault, I must change the branch asap.

Regards,
Marcelo.

PR #406, opened 2022-08-02
http: RecvMessage: Fix potential allocation failure

chouquette:
If for whatever reason the server replies without maxing out the buffer,
the code would end up doubling the buffer size nonetheless, which can
quickly cause malloc to fail, once we try to allocate 2^30 or more bytes
This was especially noticeable for large results.
Reported as https://code.videolan.org/videolan/vlc/-/issues/27173


cgisquet:
FYI, this is a realworld breakage (as indicated by the linked VLC issue): a lot of users (myself included) have been unable to view media files on their server when the client uses an up-to-date libupnp.

PR #416, opened 2022-10-05
Fix windows build documentation

mrjimenez:


Vollstrecker (approved):
I couldn't imagine anything else to say. Maybe just drop Windows and renaming Cmake Build to All Systems or something Like that could be an option, too.

Issue #424, opened 2023-03-04
memory leak in upnpapi.c ?


philippe44:
Since I'm moved to recent 1.14 (I used to be on  very old branch 1.6.19), I've observed a big memory leak at every action sent. Now, I'm very confused by this code in upnpapi.c
```
	case ACTION: {
		UpnpActionComplete *Evt = UpnpActionComplete_new();
		IXML_Document *actionResult = NULL;
		int errCode = SoapSendAction(Param->Url,
			Param->ServiceType,
			Param->Act,
			&actionResult);
		UpnpActionComplete_set_ErrCode(Evt, errCode);
		UpnpActionComplete_set_ActionRequest(Evt, Param->Act);
		UpnpActionComplete_set_ActionResult(Evt, actionResult);
		UpnpActionComplete_strcpy_CtrlUrl(Evt, Param->Url);
		Param->Fun(UPNP_CONTROL_ACTION_COMPLETE, Evt, Param->Cookie);
		free(Param);
		UpnpActionComplete_delete(Evt);
		break;
	}
```
It used to be, in previous version
```
	case ACTION: {
		struct Upnp_Action_Complete Evt;
		memset(&Evt, 0, sizeof(Evt));
		Evt.ActionResult = NULL;
		Evt.ErrCode = SoapSendAction(
			Param->Url,
			Param->ServiceType,
			Param->Act, &Evt.ActionResult);
		Evt.ActionRequest = Param->Act;
		strncpy(Evt.CtrlUrl, Param->Url, sizeof(Evt.CtrlUrl) - 1);
		Param->Fun(UPNP_CONTROL_ACTION_COMPLETE, &Evt, Param->Cookie);
		ixmlDocument_free(Evt.ActionRequest);
		ixmlDocument_free(Evt.ActionResult);
		free(Param);
		break;
	}
```
Now, I've checked the examples and neither the request noe the action are freed by the code, who receives the ACTION_COMPLETE, so it does not seem to be its duty. Obviously then this memory leaks as nobody releases it. Am I missing something obvious?

Also, here is the generated code for the ActionComplete_delete, which obviously does not free Result or Request. I don't know how it is supposed to be generated, could not find a rule in configure/makefiles
```
void UpnpActionComplete_delete(UpnpActionComplete *q)
{
	struct s_UpnpActionComplete *p = (struct s_UpnpActionComplete *)q;

	if (!p)
		return;

	p->m_ActionResult = 0;
	p->m_ActionRequest = 0;
[...]

PR #426, opened 2023-03-20
fix memory leak and ActionExAsync

philippe44:
This intends to fix https://github.com/pupnp/pupnp/issues/424. I've found at least 2 memory leaks

1- When sending Action(Ex)Async
The ActionRequest document was not released nor was the ActionResult. I initially tought that this should be done in UpnpActionComplete::UpnpActionComplete_delete() but I was wrong because ActionRequest and ActionResult do not "belong" to the Evt object. The sets methods only set a pointer and modifying generator.c is not the right solution. 
The reason is that ActionRequest "belongs" to the ACTION job and should be handled as such if we want as well that, when the job is cancelled, the memory is released as well. So I've change the free_func for ACTION to reflect that.

2- When using event subscription
The Event was not released upon Timer termination. So I've done as above, created a special free_func that handles that properly. 

In general, when something prevents the execution/submission of a job,  I've also used the *actual* free function instead of duplicating the individual free(). Note that when terminating a timer, I'm using the job's internal structure free_func() as it was already used to access arg. One can argue that it should be an opaque structure and get/set or the option to call the free function could be use instead. 

Also, I realized that ActionExAsync has never been working properly because when the job is executed, only an Action is made, never an ActionEx although the Header document is in the job's argument. I've fixed that as well and stopped there to not go too deep in that rabbit hole.

I've also removed the upnp_timeout.c/h files as they were not used anymore (and had a strange names anyway)


mrjimenez:
> Nice work @philippe44!

Indeed! :+1:

philippe44:
Thanks - I've been using it for a week no w/o issues. I'm surprised that nobody had been bitten by that before!

mrjimenez:
> Thanks - I've been using it for a week no w/o issues. I'm surprised that nobody had been bitten by that before!

Indeed, surprising. Thanks for the work.

Sorry for taking so long to do a release.

Best regards,
Marcelo.

PR #445, opened 2024-06-29
fix 32-bit format warnings

neheb:
Seems this PRIzu takes size_t, not unsigned long.

### DIFF
--- a/upnp/inc/UpnpGlobal.h
+++ b/upnp/inc/UpnpGlobal.h
@@ -44,12 +44,19 @@
 	 * PRIzx
 	 */
 	#ifdef UPNP_USE_MSVCPP
-		/* define some things the M$ VC++ doesn't know */
-		#define UPNP_INLINE _inline
+		#if _MSC_VER > 1900
+			#define UPNP_INLINE inline
+			#define PRIzd "zd"
+			#define PRIzu "zu"
+			#define PRIzx "zx"
+		#else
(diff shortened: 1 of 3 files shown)


AlaricSenat:
This is reintroducing the opposite MSVC warnings (see the CI):
```
 D:\a\pupnp\pupnp\upnp\src\genlib\net\http\httpreadwrite.c(632,7): warning C4477: 'snprintf' : format string '%lx' requires an argument of type 'unsigned long', but variadic argument 1 has type 'size_t' [D:\a\pupnp\pupnp\build\upnp\upnp_shared.vcxproj]
```

We might want to fix the `PRIzu` defined in the `UpnpGlobal` header file.

neheb:
@AlaricSenat lx is for compatibility with MSVC less than 2015?

AlaricSenat:
It's definitely for compatibility with an old MSVC version, I'm not exactly sure which one, if you could link us evidence that it's the 2015 one it would be great! :)

IMO, the correct fix would be to use `%zu` for MSVC higher than 2015 with a  `_MSC_VER` check.

neheb:
https://learn.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=msvc-170 lists improvement in format functions, but doesn't explicitly say c99 formats.

neheb:
As an aside, MinGW needs fixing too. Probably different PR.

PR #448, opened 2024-07-09
Add CORS header to avoid browser warnings

KarlStraussberger:
CORS warnings happen e.g. when using gerbera web ui with dns name

### DIFF
--- a/upnp/src/genlib/net/http/webserver.c
+++ b/upnp/src/genlib/net/http/webserver.c
@@ -1367,6 +1367,15 @@ static int process_request(
 		/*          goto error_handler; */
 		/*      } */
 	}
+	{
+		UpnpListHead *eHeader =
+			UpnpFileInfo_get_ExtraHeadersList(finfo);
+		UpnpExtraHeaders *h = UpnpExtraHeaders_new();
+		UpnpExtraHeaders_set_resp(h, "Access-Control-Allow-Origin: *");
+		UpnpListInsert(eHeader,
+			UpnpListEnd(eHeader),
+			UpnpExtraHeaders_get_node(h));
+	}
 	RespInstr->ReadSendSize = UpnpFileInfo_get_FileLength(finfo);
 	/* Check other header field. */
 	code = CheckOtherHTTPHeaders(
--- a/upnp/src/genlib/net/http/webserver.c
+++ b/upnp/src/genlib/net/http/webserver.c
@@ -231,6 +231,8 @@ static struct document_type_t gMediaTypeList[NUM_MEDIA_TYPES];
 
 /*! Global variable. A local dir which serves as webserver root. */
 membuffer gDocumentRootDir;
+/*! Global variable. A string which is set in the header field. */
+membuffer gWebserverCorsString;
 
 /*! XML document. */
 static struct xml_alias_t gAliasDoc;
@@ -475,6 +477,7 @@ int web_server_init()
 		/* decode media list */
 		media_list_init();
 		membuffer_init(&gDocumentRootDir);
+		membuffer_init(&gWebserverCorsString);
 		glob_alias_init();
 		pVirtualDirList = NULL;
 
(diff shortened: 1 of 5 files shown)


Tieske:
this will prevent browsers from sending any credentials iirc, could that cause problems?

KarlStraussberger:
> this will prevent browsers from sending any credentials iirc, could that cause problems?

If CORS kicks in the browser blocks communication to the server, including log in.

KarlStraussberger:
The browers send credentials as soon as there is a request in a script. CORS just blocks scripts from being executed. If we have a wildcard we allow scripts beeing loaded from any source.
Of course, a new library function to overwrite the value would be best, but this brings loads of version checks in applications whether the function is available or not.

Tieske:
From this link: https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSNotSupportingCredentials

What I read is that a browser cannot access any resource on the (upnp) server while providing credentials. So if there are security scheme's in effect (there are some UPnP docs on security, but again dunno the details), then with this change, those might break right?

Not saying this won't work, just wondering if there is a risk of breaking anything.

KarlStraussberger:
Wasn't aware of that restriction. As a consequence we have to add a method to set that value, which still leaves us with the problem that older binaries running with newer libraries will have to run without CORS being set at all.

upnp/src/genlib/net/http/webserver.c
```diff
@@ -1367,6 +1367,15 @@ static int process_request(
 		/*          goto error_handler; */
 		/*      } */
 	}
+	{
+		UpnpListHead *eHeader =
+			UpnpFileInfo_get_ExtraHeadersList(finfo);
+		UpnpExtraHeaders *h = UpnpExtraHeaders_new();
+		UpnpExtraHeaders_set_resp(h, "Access-Control-Allow-Origin: *");
```
randomstuff: Is this allowing *any server anywhere on the web* to access the UPnP resources? I that's the case  I don't believe that's desirable.

(I am not sure what is supposed to happen when the other `Access-Control-Allow-*` are absent).

KarlStraussberger: It's mainly controlling the access of scripts from other sources, a.k.a side-loading

KarlStraussberger:
That's completely fine. Tried to upload the files instead of creating a local branch - not a good idea.

mrjimenez:
@KarlStraussberger ,

When you feel your are done with your work, please tell me so that I can do a new release.

No hurries on my side.

Best regards,
Marcelo.

KarlStraussberger:
Will do the local branch tomorrow to make sure all is set and done.

KarlStraussberger:
@mrjimenez No more findings so you can go on as you like.

upnp/src/genlib/net/http/webserver.c
randomstuff: Won't this make it easier for a remote web server to scan your LAN (or local machine):

- for fingerprinting your LAN/device (eg. when you are using a VPN),
- for detecting the IP address, port and version of UPnP software in order to try to trigger an attack?

randomstuff:
Moved this comment here for more visibility.

Won't this make it easier for a remote web server to scan your LAN (or local machine):

    for fingerprinting your LAN/device (eg. when you are using a VPN),
    for detecting the IP address, port and version of UPnP software in order to try to trigger an attack?

KarlStraussberger:
If you add a "*" yes. That's why I revoked the change and allow to set it to values by the upnp server (e.g. gerbera).

We will allow calls from 'self' virtualHostname and externalHostname only to make sure scripts are loaded when running behind a proxy or in a container.

Vollstrecker:
> When you feel your are done with your work, please tell me so that I can do a new release.

Maybe wait with the release until I'm sure (https://github.com/amule-project/amule/issues/374) is not caused by the lib.

PR #451, opened 2024-10-03
Allow setting MaxJobsTotal for ThreadPools

### LINKED ISSUE #449: JobPool too small
Can we somehow allow increasing the pool size from 100 to say 500. I suppose that since we are hosting all web requests internally the number of allowed pending requests may be too small now.

https://github.com/gerbera/gerbera/issues/3236

KarlStraussberger:
avoids ThreadPoolAdd too many jobs: 100
in cases where heavy load is expected

Closes #449

### DIFF
--- a/upnp/inc/upnptools.h
+++ b/upnp/inc/upnptools.h
@@ -57,6 +57,13 @@ extern "C" {
 	#endif
 
 /*!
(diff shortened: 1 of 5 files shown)


KarlStraussberger:
Had to add it to the api (tools) to get the shared library straight.

PR #483, opened 2025-06-02
LinkedList.c: Use zero instead of NULL

mrjimenez:


jengelh:
(Short of using the too-modern C23 nullptr constant,) why not add #include <stddef.h> as proposed by the compiler?

mrjimenez:
Hi Jan,

That is just a personal preference. I have never seen any advantage in using NULL, so I don't use it, unless it is a requested coding style for a project.

I'll take a look at the C23 implementation to see if it brings something new to the game.

Best regards,
Marcelo.

mrjimenez:
> Another question with this is: What needs to be done to get this caught in the tests. The file is always included in the build and there's no conditionals preventing this line from being part of compilation as blame says it's 15 years in the code, how did it not fail before?

He used a different compiler, I don't know if something has changed or if I screwed up.

PR #487, opened 2025-06-10
ssdp_device.c: Fix missing casts in setsockopt()

bgilbert:
Windows thinks argument 4 of `setsockopt()` is a `const char *`.  Add casts to avoid build errors with GCC.

Fixes: #477


neheb:
it's because it is a const char*

https://learn.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-setsockopt

bgilbert:
Well, POSIX [says otherwise](https://pubs.opengroup.org/onlinepubs/9799919799/).

neheb:
right. Funny enough that documentation uses int anyway.

mrjimenez:
I don't like that this is not clearly documented and will certainly be removed by someone who commits here later. I'll create a `#define` to state the reason clearly.

PR #497, opened 2025-09-14
Tweak ssdp_device

KarlStraussberger:
Use `SO_REUSEPORT` with `SO_REUSEADDR`


KarlStraussberger:
Found the ifdef in `ssdp_server.cc` as well. Might be a problem with bitten fruits.

PR #500, opened 2025-10-13
bugfix-skip-init-on-VPN-adaptors

ss1969:
Background:
  If we call UpnpInit2(NULL, 0) on a windows system with VPN TUN mode
  enabled, this function will return -203 because it want to open the
  VPN adaptor.

Changelog:
1, Skip the VPN adaptors on windows.

Note:
1, No corresponding codes on other OS, may need to add some

Tested:
Windows 11 25H2 with Clash Verge TUN mode enabled.


mrjimenez:
Hi Ingo,

Only if you call it with the interface name equal to NULL like this `UpnpInit2(NULL, 0)`. 

You can still use a VPN interface on Windows if you name the interface explicitly.

Regards!

PR #504, opened 2025-10-23
shutdown sockets in finish routine

ss1969:
Background:
  Sockets created by accept() in schedule_request_job() will not be
shutdown in original RunMiniServer() , it only shutdown listening sockets. I tried to use a LinkedList to store sockets created in schedule_request_job(), and shutdown() them all in RunMiniServer() when stopSock is set. This will help UpnpFinish() to quit quickly.

Changlog:
1, add functions to help shutdown sockets created before.

### DIFF
--- a/upnp/src/genlib/miniserver/miniserver.c
+++ b/upnp/src/genlib/miniserver/miniserver.c
@@ -394,24 +394,12 @@ static int active_connection_cmp(void *first, void *second)
 }
 
 /*!
- * \brief Free function for active connections
- */
-static void active_connection_free(void *item)
-{
-	if (item) {
-		free(item);
-	}
-}
-
-/*!
  * \brief Add a socket to the active connections list
  */
 static void add_active_connection(SOCKET sock)
 {
 	if (!gActiveConnectionsInitialized) {
-		ListInit(&gActiveConnections,
-			active_connection_cmp,
-			active_connection_free);
+		ListInit(&gActiveConnections, active_connection_cmp, free);
 		ithread_mutex_init(&gActiveConnectionsMutex, NULL);
 		gActiveConnectionsInitialized = 1;
 	}


mrjimenez:
Hi ss1969,

Thank you for submitting the PR.

I get the general idea, but I have some concerns with a few implementation details.

In your call to `ListInit(&gActiveConnections, NULL, NULL)`, you pass two NULL parameters to the list initializaton, meaning that you will not use the comparison function and will not use the free function. But there is a contradiction when you call 

```c
			ListDelNode(&gActiveConnections, node, 1);
			free(conn);
```

The last parameter equals one, which means you want the list code to explicitly delete the item. But since you have passed NULL before, the list code will not free it. Anyway, you do free it in the following line, so, no big deal here. But you can leave the code cleaner if you supply the free function and let the list code do it.

Also, you manually coded the loop to find the socket. If you had supplied the comparison function, you could have written something like this:

```c
static int cmp_func(void *a, void *b) {
	active_connection_t *aa = a;
	active_connection_t *bb = b;

	return aa->socket == bb->socket;
}

ListInit(&gActiveConnections, cmp_func, free);
```

And then in `remove_active_connection()`:

```c
struct active_connection_t conn_sample = {
	.socket = sock
};
ListNode *node = ListFind(&gActiveConnections, NULL, conn_sample);
ListDel(&gActiveConnections, conn_sample);
```

That looks cleaner and less error prone, since you write less code and uses the list code  that, in theory, has already been well tested.

Also, in `shutdown_all_active_connections()`, I believe you have not deallocated the list items. Since we are shutting them all down, you should call `ListDestroy()` in the end of this function, otherwise, the list items will leak. Notice that you need to pass the `free` parameter to `ListInit()` to have the items freed.

Please, feel free to tell me what you think, or correct me if I made a mistake in my analysis.

ss1969:
Hi Marcelo, I totally agree with you. I've already resubmitted the PR. I wasn't careful enough before - just focused on completing the functionality. My apologies.

PR #505, opened 2025-10-30
try 'en0' when UpnpInit() with null interface under mac/ios

ss1969:
1, try "en0" when UpnpInit() with null interface.


ss1969:
Hi Marcelo, 
my original purpose was to ensure that calling UpnpInit2(Null, 0) on macOS and iOS could still correctly initialize the network card, which is why I added "en0" to the default list. So, as long as this functionality is achieved, there should be many ways to make the changes. I'm not sure what your opinion is on this. In summary, since UpnpInit2 allows a null call, it should be made as compatible as possible with more systems.

mrjimenez:
Hi @ss1969 , sorry for the delay.

I don't know what other people think, but my opinion is that it is not reasonable to put the name of the interface in the library code. There are countably infinite possibilities for that, I see no point in choosing a particular one. I think that there should be a better reason other than "MacOS will work".

PR #518, opened 2026-01-16
cmake: add missing file in Windows compilation

robUx4:
It's compiled with the autotools build but not CMake. This is only useful when compiling for older targets than Windows Vista.


Vollstrecker:
Basically correct. I'm unsure if we really should enable people running stuff (especially network related stuff) on systems that old.

mrjimenez:
Thanks for the review, @Vollstrecker !

Issue #423, opened 2023-01-12
Crash in httpreadwrite.c


AmarOk1412:
Hi

On latest release (and some previous ones), I see a crash in libupnp when downloading some Igd descriptions.

I don't have a scenario yet (I'll dig) but I have a stacktrace:

```
AddressSanitizer:DEADLYSIGNAL
=================================================================
==2769021==ERROR: AddressSanitizer: SEGV on unknown address 0x000000008000 (pc 0x7fe2bdc5ae40 bp 0x7fe21a932efc sp 0x7fe21a932e70 T230)
==2769021==The signal is caused by a READ memory access.
http_SendMessage src/genlib/net/http/httpreadwrite.c:515
http_RequestAndResponse src/genlib/net/http/httpreadwrite.c:788
http_Download src/genlib/net/http/httpreadwrite.c:897
UpnpDownloadUrlItem src/api/upnpapi.c:3500
UpnpDownloadXmlDoc src/api/upnpapi.c:3519
```

where fmt seems to be nullptr in some cases.

ingo-h:
Hi @AmarOk1412,
> I don't have a scenario yet (I'll dig)

You just use the library as normal and suddenly it crashes? How often?

AmarOk1412:
I don't think it's new. It's just that at the office, there is a looooot of devices and some NAS/Android TV announced them as a UPnP router.

I don't see this crash often, except on the office network and I think I see this ~ 2 times a week.  That's why I don't have yet a particular scenario is pretty rare and hard to get. But it seems to happen on some conditions.

ingo-h:
> where fmt seems to be nullptr in some cases.

As far as I can see from the debugger log fmt is set to 0x20000. That's not a nullptr.

AmarOk1412:
indeed but still incorrect memory causing a segv

AmarOk1412:
I reproduced the crash

It's when `num_written != buf_length`. e.g.

```
@@@@@@@@@@@@@ nw 244
@@@@@@@@@@@@@ num_written 244
@@@@@@@@@@@@@ buf_length 224
```

ingo-h:
Can you please explain what I have to do on my pupnp git clone to provoke the crash? Calling what function is essential?

ingo-h:
From the stacktrace the function `UpnpDownloadXmlDoc()` is called first. It has the declaration:

    /* \brief Downloads an XML document specified in a URL. */
    int UpnpDownloadXmlDoc(
[...]

PR #519, opened 2026-01-22
Adapt genlib and ssdp to use poll() instead of select()

moc-sfl:
select() has an upper-limit on file descriptor numbers it can monitor (see [man](https://man7.org/linux/man-pages/man2/select.2.html)). This patch switches it out for poll() and makes the associated necessary changes.

Tested with the tests under `upnp/test/`.

Related issue: https://github.com/pupnp/pupnp/issues/423

### DIFF
--- a/upnp/src/genlib/miniserver/miniserver.c
+++ b/upnp/src/genlib/miniserver/miniserver.c
@@ -63,6 +63,7 @@
 	#ifndef _WIN32
 		#include <netinet/tcp.h> /* TCP_NODELAY */
 	#endif
+	#include <poll.h>
 	#include <stdio.h>
 	#include <stdlib.h>
 	#include <string.h>
@@ -660,14 +661,19 @@ void shutdown_all_active_connections(void)
 	/*--------------------------------------------------------------------*/
 }
 
-static UPNP_INLINE void fdset_if_valid(SOCKET sock, fd_set *set)
+static UPNP_INLINE void pollfd_if_valid(
+	SOCKET sock, struct pollfd *pfd, short events)
 {
 	if (sock != INVALID_SOCKET) {
-		FD_SET(sock, set);
+		pfd->fd = sock;
+		pfd->events = events;
+		pfd->revents = 0;
+	} else {
+		pfd->fd = -1;
 	}
 }
 
-static void web_server_accept(SOCKET listen_sock, fd_set *set)
+static void web_server_accept(SOCKET listen_sock, struct pollfd *pfd)
 {
 	/*--------------------------------------------------------------------*/
 	#ifdef INTERNAL_WEB_SERVER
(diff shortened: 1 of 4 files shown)


mrjimenez:
UpnpInit2() is hanging on Windows. Here is what copilot says:

The Windows hang is caused by PR #519’s miniserver “poll()” conversion using WSAPoll() with a Linux struct pollfd layout, which breaks readiness detection on Windows and can leave the miniserver thread blocked forever. That in turn makes UpnpInit2() (called by test-upnp-init) hang.

In upnp/src/genlib/miniserver/miniserver.c at ref 7e8ffc181c6d…, the PR does this:

- On Windows it does:
#define _poll(fds, nfds, timeout) WSAPoll(fds, nfds, timeout)
- But it also defines/uses struct pollfd fds[] and POLLIN|POLLERR|POLLHUP and includes <poll.h> only on non-Windows.

That’s a fundamental mismatch: WSAPoll() expects an array of WSAPOLLFD, not POSIX struct pollfd. The fields are not guaranteed to match, and the POLL* constants also come from different headers. Result: on Windows the stop socket / listen sockets may never show revents & POLLIN, so nothing progresses and init deadlocks/hangs.

You can see the problematic part here (all in the PR version you pointed to):

```c
#ifdef _WIN32
  #include <winsock2.h>
  #include <ws2tcpip.h>
#else
  #include <poll.h>
#endif

#ifdef _WIN32
  #define _poll(fds, nfds, timeout) WSAPoll(fds, nfds, timeout)
#else
  #define _poll(fds, nfds, timeout) poll(fds, nfds, timeout)
#endif

...

struct pollfd fds[MINI_SERVER_MAX_PFDS];
...
ret = _poll(fds, nfds, -1);
```

What to request from the PR author (minimal correct fix):
They need to make the Windows path use the correct type and constants, e.g.:

- On Windows:
    - use WSAPOLLFD fds[];
    - use POLLRDNORM (or POLLIN if available via winsock headers, but typically POLLRDNORM is the safe “readable” flag in WSAPoll)
    - access pfd->revents the same way

OR (simpler / safer for pupnp):

- Keep select() on _WIN32 and only use POSIX poll() elsewhere.

mrjimenez:
@copilot resolve the merge conflicts in this pull request

mrjimenez:
Closing this in favour of #556, which contains the same core idea (replace `select()` with `poll()` for non-Windows platforms) but fixes a Windows hang introduced by this branch.

**Root cause of the hang:** `WSAPoll()` was called through the `_poll` macro with a `struct pollfd` whose `fd` field is `int` (32-bit, from MinGW's POSIX layer). On Windows x64, `WSAPoll()` expects `WSAPOLLFD` whose `fd` is `SOCKET` (64-bit). The 32-bit value passed in the `fd` slot is garbage from `WSAPoll`'s perspective, so it never reports any socket as ready — causing `UpnpInit2()` to hang indefinitely.

**Fix in #556:** `poll()` (via `_poll`) is used on non-Windows platforms only; the original `select()`-based paths are preserved on Windows. This is the minimal safe change: Linux/macOS/etc. get the `FD_SETSIZE` ceiling removed; Windows keeps the proven codepath.

#556 also adds a regression test (`test-upnp-sock-io`) that exercises `UpnpInit2()` and `UpnpSearchAsync()` to catch similar hangs in the future.

Thank you for the contribution — the `poll()` improvement itself is a good idea and the work here was the starting point for #556.

PR #534, opened 2026-03-13
Fix usage of INCLUDES DESTINATION

sge-d-o:
In ixml/CMakeLists.txt and upnp/CMakeLists.txt:
`INCLUDES DESTINATION` is not an `<artifact-kind>` and thus does not come with `<artifact-option>`s like e.g. `COMPONENT`. At least with cmake 4, the strings up to the next `<artifact-kind>` get considered as paths which do not exist, resulting in invalid `INTERFACE_INCLUDE_DIRECTORIES ...` lines in the generated files /usr/local/lib/cmake/IXML/IXML.cmake and /usr/local/lib/cmake/UPNP/UPNP.cmake.

This patch drops the false `COMPONENT UPNP_Development` lines. As a saveguard the remaining `INCLUDES DESTINATION ...` lines are moved to the end of the respective invocation of `install(TARGET...)` as the cmake manual suggests, see
https://cmake.org/cmake/help/v4.0/command/install.html#signatures.

This issue was detected when trying to build aMule with uPNP support, see https://github.com/amule-project/amule/issues/374 for further details.


Vollstrecker:
@sge-d-o Seems like I can't add you as a reviewer for #537.
This works for you?

sge-d-o:
Could add my review to #537. Unfortunately there are still issues.

PR #537, opened 2026-03-18
Fixed exports

Vollstrecker:
Fixes #534


ixml/CMakeLists.txt
```diff
@@ -51,28 +51,27 @@ if(UPNP_BUILD_SHARED)
 		PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/inc/>
 		PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src/inc/>
```
sge-d-o, line 52: The exported file /usr/local/lib/cmake/IXML/IXML.cmake still lists the non-existing directories in addition to a doubled `${_IMPORT_PREFIX}/include/upnp`:
```
71   INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include/upnp;${_IMPORT_PREFIX}/include/upnp;${_IMPORT_PREFIX}/COMPONENT;${_IMPORT_PREFIX}/UPNP_Development"
```
Probably because of a missing 
```
PUBLIC $<BUILD_INTERFACE:${PUPNP_BINARY_DIR}/>
```
as it now appears in upnp/CMakeLists.txt on line 146 of the diff.

ixml/CMakeLists.txt
sge-d-o, line 50: Just noticed the difference to upnp/CMakeLists.txt, line 143 below. May be this contributes to my observation described with line 52 of the diff.

sge-d-o (changes requested):
The issue seems to have been fixed for /usr/local/lib/cmake/UPNP/UPNP.cmake while /usr/local/lib/cmake/IXML/IXML.cmake still is faulty.

ixml/CMakeLists.txt
Vollstrecker: I missed the bogus INCLUDES for ixml-static. Problem was just, that includes wasn't recognized and everything until the next recognized part will be addedm like COMPONENTS.

I now see only

'INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include/upnp"'

ixml/CMakeLists.txt
Vollstrecker: Nope, PRIVATE indicates in this case that this will not be part of the exported interface. That's also the reason why this is not classified as BUILD_ or INSTALL_INTERFACE as it only counts for building the target itself.

sge-d-o (approved):
Looks good to me. Thanks.

PR #550, opened 2026-05-20
High memory usage after #504

RyanHope97:
#504 added a feature to store sockets in a list to improve shutdown performance. Memory is allocated for every connection, but as far as I can tell it is only freed on shutdown. This results in memory usage slowly increasing over time.

I was able to reproduce this with the tv_device sample.

I've added the appropriate free call after a request has been handled, but I'm not familiar enough with the library architecture to know if this is the most idiomatic fix.

Below are some valgrind massif heap profiler screenshots.

Before
Screenshot 2026-05-20 162134

After
Screenshot 2026-05-20 162149

### DIFF
--- a/upnp/src/genlib/miniserver/miniserver.c
+++ b/upnp/src/genlib/miniserver/miniserver.c
@@ -533,6 +533,7 @@ error_handler:
 		}
 		handle_error(&info, http_error_code, major, minor);
 	}
+	remove_active_connection(info.socket);
 	sock_destroy(&info, SD_BOTH);
 	httpmsg_destroy(h_msg);
 	free(request);


mrjimenez (approved):
The fix is correct. `free_handle_request_arg` is the thread pool's cancellation callback (for jobs cancelled before execution), not the normal completion path — so `remove_active_connection` was never called after a request completed normally. Adding it here, before `sock_destroy`, is the right place.

One pre-existing gap worth noting: `upnp/src/genlib/miniserver/miniserver.c`, in `handle_request`, the early-return path when `sock_init_with_ip` fails doesn't call `remove_active_connection`:

```c
if (ret_code != UPNP_E_SUCCESS) {
    free(request);
    httpmsg_destroy(h_msg);
    return;   /* connection added in schedule_request_job is never removed */
}
```

The connection was already added to `gActiveConnections` before this function runs. Adding `remove_active_connection(sock)` here (the `sock` local is already saved before the call) would close that gap. Not a blocker — just the same root cause surviving in a less-travelled path.

PR #579, opened 2026-05-25
fix: bind SSDP sockets to selected interface address (issue #227)

### LINKED ISSUE #227: SSDP binds to INADDR_ANY
There is a lot of logic to find suitable IPv4/v6 addresses, but by the looks of it that is ignored by SSDP server code which just binds with `INADDR_ANY` regardless?

https://github.com/pupnp/pupnp/blob/643f9c3ea21b0312865e9e8ed49653850cbd3a55/upnp/src/ssdp/ssdp_server.c#L973-L977

Similar code exists for IPv6 too.

I am not a sockets programming expert, but looks odd to me?

mrjimenez:
Fixes #227

## What

`create_ssdp_sock_v4()`, `create_ssdp_sock_v6()`, and
`create_ssdp_sock_v6_ula_gua()` were binding to `INADDR_ANY` /
`in6addr_any` despite `gIF_IPV4` / `gIF_IPV6` / `gIF_IPV6_ULA_GUA`
already being available. On a multi-homed host this caused the SSDP
server to accept traffic on all interfaces, not just the one selected
at `UpnpInit2()` time.

## Fix

Bind to the specific interface address at all three sites. Multicast
delivery is unaffected: the `IP_ADD_MEMBERSHIP` / `IPV6_JOIN_GROUP`
joins already scope multicast reception to the correct interface.

Also updates the `bind()` error log in `create_ssdp_sock_v4()` and
`create_ssdp_sock_v6_ula_gua()` to print the actual address instead of
a hardcoded `INADDR_ANY` literal.

## POC Tests

`upnp/test/test_ssdp_bind_specific.c` — verifies (on Linux) that after
`UpnpInit2()` the SSDP IPv4 server socket is bound to the specific
interface address, not `INADDR_ANY`. Uses `/proc/self/fd` inode
cross-referencing against `/proc/net/udp` to identify the library's
own socket. Skipped on non-Linux platforms.

## Checklist
- [x] Fix implemented
- [x] POC test fails on unfixed code, passes on fixed code (TDD confirmed locally)
- [x] No regressions in existing test suite
- [x] clang-format clean
- [x] CI green

### DIFF
--- a/upnp/src/ssdp/ssdp_server.c
+++ b/upnp/src/ssdp/ssdp_server.c
@@ -968,7 +968,7 @@ static int create_ssdp_sock_v4(
 	#endif /* BSD, __APPLE__ */
 	memset(&__ss, 0, sizeof(__ss));
 	ssdpAddr4->sin_family = (sa_family_t)AF_INET;
-	ssdpAddr4->sin_addr.s_addr = htonl(INADDR_ANY);
+	inet_pton(AF_INET, gIF_IPV4, &ssdpAddr4->sin_addr);
 	ssdpAddr4->sin_port = htons(SSDP_PORT);
 	ret = bind(*ssdpSock, (struct sockaddr *)ssdpAddr4, sizeof(*ssdpAddr4));
 	if (ret == -1) {
@@ -977,8 +977,8 @@ static int create_ssdp_sock_v4(
 			SSDP,
 			__FILE__,
 			__LINE__,
-			"Error in bind(), addr=0x%08X, port=%d: %s\n",
-			INADDR_ANY,
+			"Error in bind(), addr=%s, port=%d: %s\n",
+			gIF_IPV4,
 			SSDP_PORT,
 			errorBuffer);
 		ret = UPNP_E_SOCKET_BIND;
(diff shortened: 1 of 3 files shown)


KarlStraussberger:
After upgrading my local branch, gerbera was not discovered by any upnp client. Searching for some reason, I found https://stackoverflow.com/questions/74940771/not-receiving-broadcast-udp-messages-when-binding-to-specific-address

Seems the assumption was wrong that binding to the ip solves multicast issues.

mrjimenez:
This PR introduced a regression that breaks UPnP device discovery entirely.

**Root cause:** binding the SSDP server sockets to the specific interface unicast address (e.g. `192.168.x.y:1900`) prevents the Linux kernel from delivering multicast datagrams to them. All SSDP M-SEARCH queries are sent to `239.255.255.250:1900`; a socket bound to a unicast address never receives them, regardless of `IP_ADD_MEMBERSHIP`. The assumption in the PR description that *"binding to a unicast address does not suppress delivery of multicast datagrams the socket has joined"* is incorrect on Linux.

Multicast interface scoping was already correctly handled by `IP_ADD_MEMBERSHIP` with `imr_interface = gIF_IPV4`, so the `INADDR_ANY` bind was necessary and correct.

A fix is in PR #601. It reverts the three bind sites to `INADDR_ANY` / `in6addr_any` and replaces the flawed `test_ssdp_bind_specific` with a behavioral test (`test_ssdp_multicast_recv`) that sends a real M-SEARCH and verifies the library responds.

Thank you to @KarlStraussberger for the quick report.

PR #584, opened 2026-05-27
ci: install pytest before ctest so Python tests run on Ubuntu runners

### LINKED ISSUE #583: Install pytest in CI so Python tests are not silently skipped
## Problem

The Python regression tests registered via `tests/CMakeLists.txt` are silently skipped on every GitHub Actions run because `pytest` is not installed on the runners.

The affected tests are:
- `test-init-py`
- `test-issue-110-py`
- `test-issue-250-py` (added in #582)

When CMake detects that `pytest` is absent it prints:

```
-- pytest not found; skipping Python tests
```

and returns early, so `ctest` never registers those test targets. The CI log shows no failures — but also no runs — giving a false sense of coverage.

## Root cause

The CI workflow (`build.yml`) does not install `pytest` before invoking `ctest`. The Ubuntu runner ships Python 3 but not `pytest`.

## Suggested fix

Add a `pip install pytest` step to the CI workflow before the build/test step, for example:

```yaml
- name: Install Python test dependencies
  run: pip install pytest
```

This should be applied to at least the `ubuntu-latest` job. The Windows and macOS jobs may already have `pytest` or can be updated similarly.

## Impact

Without this fix, any regressions caught by the Python test suite will go undetected in CI. The tests are effective locally (they caught the issue #250 DoS scenario) but provide no CI-level safety net.

mrjimenez:
## Summary

- Adds a `pip install pytest` step to the `linux` job in `ccpp.yml` so that `test-init-py`, `test-issue-110-py`, and `test-issue-250-py` are registered and executed by `ctest` instead of being silently skipped.
- Adds the same step (gated on `matrix.tool == 'cmake'`) to the `build_asan` job, which also runs `ctest`.

## Test plan

- [ ] Confirm that CI output for the `linux` job shows all three Python test targets running (not skipped).
- [ ] Confirm that the `build_asan` (cmake) job likewise runs the Python tests.
- [ ] Confirm no regressions on Windows/macOS/OmniOS jobs (unaffected).

Fixes #583

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -68,6 +68,8 @@ jobs:
       fail-fast: false
     steps:
       - uses: actions/checkout@v6
+      - name: Install Python test dependencies
+        run: pip install pytest
       - name: Configure (create build system)
         run: |
           cmake \
@@ -220,6 +222,9 @@ jobs:
         run: |
           ./bootstrap && \
           ./configure --enable-debug ${{ env.sanitizer_autotools_flags }}
+      - name: Install Python test dependencies
+        if: ${{ matrix.tool == 'cmake' }}
+        run: pip install pytest
       - name: configure (cmake)
         if: ${{ matrix.tool == 'cmake' }}
         run: |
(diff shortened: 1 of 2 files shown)


mrjimenez (commented):
The `ccpp.yml` change is correct and working — the three Python test targets now actually run in the sanitizer (cmake) job instead of being silently skipped. The CI failure it exposed is pre-existing: Python's interpreter intentionally skips freeing certain allocations at shutdown, which AddressSanitizer's leak detector treats as failures.

`tests/CMakeLists.txt` already preloads `libasan.so` when building with ASAN, but was missing `ASAN_OPTIONS=detect_leaks=0` to suppress Python's own interpreter leaks. Fixed in d57b93ed.

PR #587, opened 2026-05-28
switch to use genex for sources and headers

Vollstrecker:
When genex are used we can get rid of all that if's

### DIFF
--- a/upnp/CMakeLists.txt
+++ b/upnp/CMakeLists.txt
@@ -37,47 +37,31 @@ set(UPNP_SOURCES
 	src/threadutil/ThreadPool.c
 	src/threadutil/TimerThread.c
 	src/urlconfig/urlconfig.c
+	$<$<BOOL:${WIN32}>:src/inet_pton.c>
+	$<$<BOOL:${UPNP_ENABLE_GENA}>:src/gena/gena_callback2.c>
+	$<$<BOOL:${UPNP_ENABLE_GENA}>:src/gena/gena_ctrlpt.c>
+	$<$<BOOL:${UPNP_ENABLE_GENA}>:src/gena/gena_device.c>
+	$<$<BOOL:${UPNP_ENABLE_SOAP}>:src/soap/soap_common.c>
+	$<$<BOOL:${UPNP_ENABLE_SOAP}>:src/soap/soap_ctrlpt.c>
+	$<$<BOOL:${UPNP_ENABLE_SOAP}>:src/soap/soap_device.c>
+	$<$<BOOL:${UPNP_ENABLE_SSDP}>:src/ssdp/SSDPResultData.c>
+	$<$<BOOL:${UPNP_ENABLE_SSDP}>:src/ssdp/SSDPResultDataCallback.c>
+	$<$<BOOL:${UPNP_ENABLE_SSDP}>:src/ssdp/ssdp_device.c>
+	$<$<BOOL:${UPNP_ENABLE_SSDP}>:src/ssdp/ssdp_ctrlpt.c>
+	$<$<BOOL:${UPNP_ENABLE_SSDP}>:src/ssdp/ssdp_server.c>
+	$<$<OR:$<BOOL:${UPNP_ENABLE_GENA}>,
+		$<BOOL:${UPNP_ENABLE_OPTIONAL_SSDP_HEADERS}>>:
+		src/uuid/md5.c>
+	$<$<OR:$<BOOL:${UPNP_ENABLE_GENA}>,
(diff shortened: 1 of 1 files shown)


upnp/CMakeLists.txt
```diff
@@ -98,14 +82,9 @@ list(
 	inc/UpnpSubscriptionRequest.h
 	inc/UpnpUniStd.h
 	${PUPNP_BINARY_DIR}/upnp/inc/upnpconfig.h
+	$<$<BOOL:${UPNP_ENABLE_HELPER_API_TOOLS}>:${PUPNP_SOURCE_DIR}/upnp/inc/upnptools.h>
```
mrjimenez, line 85: Minor: all other headers in this list use relative paths (`inc/Callback.h`, `inc/upnp.h`, etc.), but `upnptools.h` switches to an absolute path here. The original `list(APPEND UPNP_HEADERS inc/upnptools.h)` also used a relative path. These resolve to the same file, but the inconsistency is unexplained.

Suggested fix:
```cmake
$<$<BOOL:${UPNP_ENABLE_HELPER_API_TOOLS}>:inc/upnptools.h>
```

upnp/CMakeLists.txt
```diff
@@ -37,47 +37,31 @@ set(UPNP_SOURCES
...
+	$<$<BOOL:${UPNP_ENABLE_GENA}>:src/gena/gena_device.c>
+	$<$<BOOL:${UPNP_ENABLE_SOAP}>:src/soap/soap_common.c>
+	$<$<BOOL:${UPNP_ENABLE_SOAP}>:src/soap/soap_ctrlpt.c>
+	$<$<BOOL:${UPNP_ENABLE_SOAP}>:src/soap/soap_device.c>
+	$<$<BOOL:${UPNP_ENABLE_SSDP}>:src/ssdp/SSDPResultData.c>
+	$<$<BOOL:${UPNP_ENABLE_SSDP}>:src/ssdp/SSDPResultDataCallback.c>
+	$<$<BOOL:${UPNP_ENABLE_SSDP}>:src/ssdp/ssdp_device.c>
+	$<$<BOOL:${UPNP_ENABLE_SSDP}>:src/ssdp/ssdp_ctrlpt.c>
+	$<$<BOOL:${UPNP_ENABLE_SSDP}>:src/ssdp/ssdp_server.c>
+	$<$<OR:$<BOOL:${UPNP_ENABLE_GENA}>,
```
mrjimenez, line 52: Nit: the same `$<$<OR:$<BOOL:${UPNP_ENABLE_GENA}>,$<BOOL:${UPNP_ENABLE_OPTIONAL_SSDP_HEADERS}>>` condition is repeated verbatim three times, and the multi-line layout makes it look (at a glance) like the file paths have leading whitespace. Introducing a helper variable would eliminate the repetition and make each source line self-evident:

```cmake
set(_UUID_COND "$<OR:$<BOOL:${UPNP_ENABLE_GENA}>,$<BOOL:${UPNP_ENABLE_OPTIONAL_SSDP_HEADERS}>>")
...
    $<${_UUID_COND}:src/uuid/md5.c>
    $<${_UUID_COND}:src/uuid/sysdep.c>
    $<${_UUID_COND}:src/uuid/uuid.c>
```

This is purely cosmetic — the current form is functionally correct (CI confirms it).

mrjimenez (approved):
CI passes cleanly across all platforms. Two minor observations below.

upnp/CMakeLists.txt
Vollstrecker: Nope. the other inc/... are expanded at configure-stage while the genex is evaluated a generation-stage in BINARY_DIR resulting in trying to install from there.

upnp/CMakeLists.txt
Vollstrecker: I think this will make them even more hard to read. Many people struggle with genex anyways, placing a var with a genex in a genex doesn't help them (especially as they would need to scroll to see the var).

I could split them into separate checks so they fit on one line again. Adding the file twice is no problem.

upnp/CMakeLists.txt
mrjimenez: You are right — I had not considered the generation-stage resolution against BINARY_DIR. The absolute path is the right call. Thanks for the explanation.

upnp/CMakeLists.txt
mrjimenez: Makes sense. Do you plan to push the single-line variant, or would you prefer to merge as-is? Happy to proceed either way, just let us know.

Vollstrecker: That's your decision, I'm fine with both.

PR #594, opened 2026-05-30
Components

Vollstrecker:
Here is a little rework to enable users to do find_package(UPNP COMPENTS Shared) so they don't get both targets. I'm a bit dissappointed that noone so far notice that pthreads is used but never checked in the exports.

While doing this I also fixed CmDaB so after this is merged I'm sure I can try another PR to migrate to the new version.

I also noticed that the new test runs on shell, so I also will do another PR and rewrite that one in CMake, so it runs everywhere.

Also in a next PR I would remove the deprecation-warning and i still need the decision if it should error to really let User know that they use the old one for 1 or 2 versions, or if the old ones should just be silently ignored.

When doing this I also would switch to UPNP_BUILD_TESTING (with a notice if BUILD_TESTING is set and defaulting to this), as BUILD_TESTING is the last not personalized. For that I would like to know if that one would be enough, or if ixml shouzld better get it's own.


mrjimenez (changes requested):
Thank you for this PR — component support for `find_package(UPNP COMPONENTS Shared/Static)` is a real gap and the overall structure is good. A few bugs need fixing before this can merge:

**Blocking issues:**

1. **Fixture name typo — `add_subdirectory_build` tests run out of order.**
   In `ixml/test/CMakeLists.txt`, `ixml_add_subdirectory_build` declares `FIXTURES_REQUIRED as_config` but the setup fixture is named `ixml_as_config`. Same problem in `upnp/test/CMakeLists.txt` (`as_config` → `upnp_as_config`). On a parallel `ctest` run the build step can start before configure finishes.

2. **Copy-paste bug — static targets are never tested.**
   Every test template (`add_subdirectory.cmake.in`, `add_subdirectory_exclude.cmake.in`, `find_package_no_components.cmake.in`, `find_package_test.cmake.in` — both ixml and upnp) has the second `if(UPNP_BUILD_SHARED)` block guarding the Static target. That condition should be `if(UPNP_BUILD_STATIC)`. Eight files affected.

3. **`find_package_wrong_components.cmake.in` — missing closing `)` on `find_package` call.**
   Both `ixml/test/` and `upnp/test/` versions. CMake emits a parse error instead of the intended "Unknown component" error. Since `WILL_FAIL TRUE` is set the test still passes, but it's testing a syntax error rather than the component-validation logic. Add the closing `)`.

4. **Wrong install prefix in `upnp_find_package_wrong_components_configure`.**
   `-DCMAKE_INSTALL_PREFIX=.../UPNP_test_install` (capital UPNP) doesn't match the directory installed by the `upnp_install` fixture (`upnp_test_install`, lowercase). The wrong-components test will fail to locate the installed package.

**Should also fix:**

5. **`CMAKE_CXX_STANDARD 14` was removed but not re-added.**
   `set(CMAKE_CXX_STANDARD 14)` and `set(CMAKE_CXX_STANDARD_REQUIRED ON)` were dropped from the top-level preamble. The project has `.cpp` test files that may depend on C++14 behaviour. Please re-add these after `enable_language(CXX)` inside the `if(BUILD_TESTING)` block, or set them as target properties.

6. **`upnp/test/find_package_test.cmake.in` has the wrong project name** (`ixml_add_find_package_test` → `upnp_add_find_package_test`).

Minor: new `.cmake.in` files are missing trailing newlines; `MAKE_DIRECTORY` names don't match the `configure_file` targets for the no-components and wrong-components cases.

Vollstrecker:
That was fast. Everything fixed except the CXX_STANDARD. This was added for gtest and there is only an empty skeleton in there. Restricting the standard because of this is unneeded. Imho the complete block is useless and just prolonges the build when gtest is built, too.

The OpenSSL thing will be fixed after dining.

mrjimenez:
Good, looking forward to the OpenSSL fix — that's the one causing failures on all platforms. The installed `UPNP.cmake` exports `OpenSSL::SSL` in the link interface but `UPNPConfig.cmake` never calls `find_dependency(OpenSSL)` in the consumer scope, so `set_target_properties` fails when the consumer tries to configure. Adding this to `UPNP.cmake.in` should do it, what do you think?

```cmake
if(@UPNP_ENABLE_OPEN_SSL@)
    find_dependency(OpenSSL)
endif()
```

One more finding from the CI logs while you're at it: the TSan build fails for a different reason. The integration test `add_test` configure commands forward `-DCMAKE_C_FLAGS=${CMAKE_C_FLAGS}` to child projects. In the TSan build that flag contains `-fsanitize=thread`, which causes CMake's own compiler detection in the child project to segfault (it compiles a test binary with the sanitizer flag but without linking the TSan runtime). Since these tests are checking CMake integration rather than code correctness, they don't need to inherit the parent's sanitizer flags — removing `-DCMAKE_C_FLAGS=${CMAKE_C_FLAGS}` from all the `add_test` configure commands in both `ixml/test/CMakeLists.txt` and `upnp/test/CMakeLists.txt` should fix the TSan job.

mrjimenez:
Two more things from the latest CI run:

**Incomplete fixture fix**

The fixture name was corrected in `ixml/test/CMakeLists.txt` but `upnp/test/CMakeLists.txt` still has `FIXTURES_REQUIRED as_config` on `upnp_add_subdirectory_build` — it should be `upnp_as_config`.

**`add_subdirectory` tests: consider removing them**

The `ixml_add_subdirectory_build` and `upnp_add_subdirectory_build` tests are timing out because they rebuild all of pupnp from scratch inside a child project — and with `DOWNLOAD_AND_BUILD_DEPS=ON` they also pull and build GTest, CmDaB, etc. That easily blows past ctest's 30-second timeout on slower runners (Windows, MSys, OmniOS).

More importantly, the `add_subdirectory` scenario doesn't need dedicated integration tests: the CI build itself *is* an `add_subdirectory` scenario from CMake's perspective, so this coverage already exists for free. The `find_package` tests, on the other hand, test something genuinely new — that the *installed* config files (`UPNPConfig.cmake`, `IXMLConfig.cmake`) work correctly for a consumer who only has the installed artifacts. That's the real value of this PR and those tests are well worth keeping.

My suggestion: drop the `add_subdirectory_test` and `add_subdirectory_exclude_test` blocks (and their corresponding `.cmake.in` templates) from both `ixml/test/CMakeLists.txt` and `upnp/test/CMakeLists.txt`. If an `add_subdirectory` integration test is wanted in the future, it should at minimum pass `-DBUILD_TESTING=OFF -DDOWNLOAD_AND_BUILD_DEPS=OFF` to avoid the full rebuild cost. What do you think?

Vollstrecker:
> Two more things from the latest CI run:
 
> The fixture name was corrected in `ixml/test/CMakeLists.txt` but `upnp/test/CMakeLists.txt` still has `FIXTURES_REQUIRED as_config` on `upnp_add_subdirectory_build` — it should be `upnp_as_config`.

I suspected.
 
> **`add_subdirectory` tests: consider removing them**

Maybe one of them is enough.

> The `ixml_add_subdirectory_build` and `upnp_add_subdirectory_build` tests are timing out because they rebuild all of pupnp from scratch inside a child project — and with `DOWNLOAD_AND_BUILD_DEPS=ON` they also pull and build GTest, CmDaB, etc. That easily blows past ctest's 30-second timeout on slower runners (Windows, MSys, OmniOS).

The 30 seconds isn't really needed. 

> More importantly, the `add_subdirectory` scenario doesn't need dedicated integration tests: the CI build itself _is_ an `add_subdirectory` scenario from CMake's perspective, so this coverage already exists for free.

Not really, The CI just builds the source, for example the tests use upnp_shared for linking, The add_subdirectory test make sure the same targets are used no matter if the exported config is used, or the the source is included as chainbuild. For sure it's debatable if the exclude_flag makes adifference.

> My suggestion: drop the `add_subdirectory_test` and `add_subdirectory_exclude_test` blocks (and their corresponding `.cmake.in` templates) from both `ixml/test/CMakeLists.txt` and `upnp/test/CMakeLists.txt`. If an `add_subdirectory` integration test is wanted in the future, it should at minimum pass `-DBUILD_TESTING=OFF -DDOWNLOAD_AND_BUILD_DEPS=OFF` to avoid the full rebuild cost. What do you think?

I didn't check the BUILD_TESTING default, turning that of will cut build-time a good amount, including not building gtest, that should help a lot.

However, this all doesn't help with asan still bleeds in somehow and tsan segfaulted.

Vollstrecker:
> it should at minimum pass `-DBUILD_TESTING=OFF -DDOWNLOAD_AND_BUILD_DEPS=OFF` to avoid the full rebuild cost. What do you think?

Oh, and this would break the Windows-build as pthreads4win aren't there.

mrjimenez:
Thanks for the diagnosis — you were right that ASan is still bleeding in, but digging into the logs shows the mechanism is different from what any of us assumed. Here's what's actually happening in each case.

---

**ASan (cmake sanitizer build) — you were right, but the path is different**

I originally said the fix was to remove `-DCMAKE_C_FLAGS` from the child project configure commands, and you did that. But the failure is not about flags propagation at all.

The configure step (`ixml_find_package_configure`) now **passes**. The build step (`ixml_find_package_build`) fails at link time, on the static target only:

```
/usr/bin/ld: .../ixml_test_install/lib/libixml.a(document.c.o):
  undefined reference to `__asan_report_store8'
/usr/bin/ld: undefined reference to `__asan_stack_malloc_0'
... (many more)
```

The installed `libixml.a` was compiled with `-fsanitize=address`, so its object files are full of `__asan_*` references. The child project links against it without ASan, so those symbols are unresolved. The shared target (`test_document_shared`) links fine — `libasan.so` is already baked into the `.so`. Only the static link breaks.

This is a structural incompatibility: a static library compiled with ASan cannot be linked by a project that isn't also using ASan. No amount of flag-stripping fixes it. The integration test static targets simply can't work inside a sanitizer build as currently structured. The options are: skip the static target tests in sanitizer builds (add `ENVIRONMENT` or a `SKIP_RETURN_CODE` condition), or accept that these tests will be excluded from sanitizer CI.

---

**TSan — the root cause is `LD_PRELOAD`, not flags**

I told you to remove `-DCMAKE_C_FLAGS` and you did that. But the TSan build is still segfaulting at compiler detection in the child projects. The logs confirm the crash is still happening in `findpackage_test_build`, not in the flag stripping.

The actual cause: the TSan CI job wraps the entire ctest invocation with:
```
LD_PRELOAD=$(gcc -print-file-name=libtsan.so) ctest ...
```
Every process ctest spawns inherits that `LD_PRELOAD` — including the `cmake` configure commands the integration tests run as child processes. When cmake tries to compile a tiny test binary to detect the C compiler, that binary has `libtsan.so` force-loaded without a properly initialised TSan runtime, and segfaults.

The fix is already used elsewhere in the codebase for the upnp tests:
```cmake
set_tests_properties(ixml_find_package_configure
    PROPERTIES ENVIRONMENT "LD_PRELOAD=")
```
Adding `ENVIRONMENT "LD_PRELOAD="` to all integration test configure steps (both ixml and upnp, for all four scenarios) clears the preload before cmake is invoked and lets the compiler detection complete normally.

Sorry for the misdirection on the TSan side — the flag removal was the right instinct but the wrong target.

Vollstrecker:
k, I now added the options rework here while adding options for deactivating ixml/upnp tests also disabling the integrations tests separate. For the asan, tsan, Win and Msys builds they are deactivated. I expected OmniOS would also be too slow, but with the longest one running 19.65 seconds there's enough headroom.

While watching the logs I've seen that we can deactivate shared or static, but building only ixml isn't possibe, maybe we should think about that later.

If you're happy now, this should be done.

mrjimenez:
Man, this is great, much cleaner than what the AI suggested.

Yeah, we should consider building ixml and upnp separeted, but I guess this is usefull for testing only. I don't know anyone that uses ixml.

Thanks for the work you put into this.

Vollstrecker:
Yes, that's why I stay on handwriting stuff - beside the fact that I think it's important to fully understand what you do (don't ask about autoheader.cmake, I understood while writing).

I've always been unsure if ixml is used somehwere on it's own. That's why I designed that whole thing to be separatable as it's own project, but over time I more got the feeling that merging all into one lib would also be an option.

mrjimenez (approved):

Vollstrecker:
I know I could, but somehow it feels strange merging into your project.

Forgot to mention: My problem with AI is that they want to please you. They try to make working whatever you show them instead if honestly saying "You started that the wrong way, throw that crap away and start over clean".

mrjimenez:
The problem was that you made an intermediate merge to main, GitHub was unable to do both a manual merge and a rebase merge. I had to do a squash merge, which I usually don't like to do because it hides the work done. But I was affraid to make CI red again if I tried to linearize it manually.

As for the AI pleasing stuff, I know that, you need to have personality. When AI starts agreeing with you, that is a trigger that you should not let go and instead go deep into the matter. That happens **a lot**. I saw somewhere that the new 4.8 Anthropic model has been classified as more "honest" :rofl: meaning it will not flatter you so much.

PR #597, opened 2026-05-31
fix: restore UPNP_ENABLE_DEBUG option in cmake/options.cmake

mrjimenez:
## Summary

- `UPNP_ENABLE_DEBUG` was declared as an `option()` on `main` but the declaration was lost when PR #594 removed the deprecated option wrapper machinery
- The `set(UPNP_HAVE_DEBUG ${UPNP_ENABLE_DEBUG})` mapping survived, so builds were not broken — debug was just silently always OFF with no way for users to enable it via `-DUPNP_ENABLE_DEBUG=ON`
- This restores the single missing line in its correct alphabetical position

## Test plan

- [x] Confirm `cmake -DUPNP_ENABLE_DEBUG=ON` is accepted and propagates to `upnpconfig.h`
- [x] CI green


Vollstrecker:
Great. When I pushedthe option it show me the whole Block as conflicting and I wasn't able to see that on option in the middle and wondered what happened. Rebasing would have been an option, but that's something the gui doesn't offer in that area.

PR #605, opened 2026-06-21
build: complete migration to CMake-only build system

mrjimenez:
## Summary

- Removes all Autotools infrastructure: `configure.ac`, `bootstrap`, all `Makefile.am` files, `m4/` macros, `cmake/autoheader.cmake`, and `scripts/comp-autotools.sh`
- Version numbers are now defined directly in `CMakeLists.txt`; the old `cmake/autoheader.cmake` that parsed `configure.ac` at configure time is replaced by a static `cmake/autoconfig.h.in` template
- Source tarball generation moves from `make distcheck` to `cpack -G TBZ2 --config build/CPackSourceConfig.cmake`
- CI workflows, `scripts/home_keeping.sh`, and `README.md` updated to match
- Adds `docs/cmake-migration.md` documenting everything that changed
- Adds `scripts/test_release_workflow.sh` to simulate the full release workflow locally for regression testing

## Test plan

- [ ] `scripts/test_release_workflow.sh` passes locally (all 6 steps, clean working tree after exit)
- [ ] CI passes on this PR (sanitizer build, OmniOS, full matrix)
- [ ] `cmake -B build && cmake --build build && ctest --test-dir build` passes


Vollstrecker:
One step closer to world-domination.

I like that you mention that historically autotools and cmake have been supported. It feels like 2 but I guess it's jusgt around 5 years.

Little remark, iirc (didn't check as I'm on the road again) ixml and upnp are set as project in their dir, so the so-version in ixml could be IXML_VERSION instead of constructing it by hand (same for upnp). Doesn't make a difference but save ~ 60 bytes.

mrjimenez:
```bash
CMAKE_MINIMUM_REQUIRED(VERSION 3.11)
CMAKE_POLICY(VERSION 3.13)

include(cmake/CmDaB.cmake)
include(cmake/test-functions.cmake)

# 1. Source of Truth for IXML
set(IXML_VERSION_MAJOR 11)
set(IXML_VERSION_MINOR 1)
set(IXML_VERSION_PATCH 8)

# 2. Source of Truth for UPNP
set(UPNP_VERSION_MAJOR 22)
set(UPNP_VERSION_MINOR 0)
set(UPNP_VERSION_PATCH 3)

# 3. Calculate Umbrella Version
math(EXPR PUPNP_VERSION_MAJOR "${IXML_VERSION_MAJOR} + ${UPNP_VERSION_MAJOR}")
math(EXPR PUPNP_VERSION_MINOR "${IXML_VERSION_MINOR} + ${UPNP_VERSION_MINOR}")
math(EXPR PUPNP_VERSION_PATCH "${IXML_VERSION_PATCH} + ${UPNP_VERSION_PATCH}")
set(PUPNP_VERSION_STRING "${PUPNP_VERSION_MAJOR}.${PUPNP_VERSION_MINOR}.${PUPNP_VERSION_PATCH}")

project(PUPNP
	VERSION ${PUPNP_VERSION_STRING}
	LANGUAGES C)
```
The only catch is that libupnp next version is going to be 33.1.11 :D

Vollstrecker:
Then I'll have to take a closer look at wednesday.

mrjimenez:
Great, have a look and tell me what you think.

I have not come up with something better to cope with two libraries at the same time. The problem is that libupnp and libixml should be two different projects, but separating them now would be a hell of a headache. Adopting semantic versioning and adding the release numbers seemed like the most sensible thing to do.

Vollstrecker:
Sorry for the delay.

I thought a bit about all that. Didn't implement as the options are too different.

We could:

1: use the current project-call (they are e.g. ixml/CMakeLists.txt and use the version-vars from there to save some bytes. Changes basically only maintenance.

2: Change the calculated semver to date-based like pupnp-20260628
- Pros: The numbers don't get as big as now
- Cons: Users would need to adjust

3: Bump ixml version to match upnp and use only that one for evrything
- Pros
    - No calculation of Tags to push
    -  Version are not that big
- Cons:
    - User would need to adjust
    -  a fix in ixml changes version of upnp and vice-versa, especially not wanted if one needs an so-version bump

4: Split them into separate projects. Both are alsmost prepared for that.
- Pros:
    - No version calculation
    - No influence on each other
- Const:
    - You know: Users
    - upnp would need to search for ixml instead of assuming it's available
    - If there's a major version change in imxl we would need to decide of upnp only supports the latest version, or maybe there will be some #ifdef for the version.

5: Merge them and ship them as pupnp containing ixml and upnp. We could still provide both as components UPNP::upnp-shared/static and UPNP::ixml-shared/static
- Pros:
    - No version calculation
    - No lib search
    - No thinking about what happens with incompatible changes on ixml.
    - Biggest impact with least code-changes
- Cons:
    - As always Users
    - Almost the same as  bumping the ixml-version to have only one.

mrjimenez:
But what are the cons of the current (new) version scheema? I think it is the best of all worlds, given we have two libraries to deal. The only con I see is that the number got large. Is that a problem? It is not like it will get into three digits so soon, and we are already at two digits in the minor.

In my opinion, the project version calculation is not a bad thing, it is the solution.

Ixml as a single separate library does not make much sense, I don't think anyone would use it. There is the infamous DOMString implementation that is not standard compliant. As far as I know, only libupnp uses it.

Date based naming is bad, it communicates nothing.

- Pros:
  - We have a single source of truth for all library numbers in one file in the root of the project.
  - That single source of truth is per library, one numbering does not interfere with the other.
  - We get real semver for each library.
  - We get real semver for the project.
  - No more SO_NAME regexp madness.
  - Major number starts acting like a major.
  - Minor number stops behaving like a major, which added confusion to the whole numbering.
  - No worries about the project version, it is automatically calculated.
  - No lib search, Ixml is always available, there is no dependency problem.
  - Releasing new versions is made much easier, restricted to changing basically the root CMakeLists.txt.
  - Users won't notice since nothing actually changed besides the larger major.

- Cons:
  - A slightly larger major in the project number. But the same amount of digits of our old minor.

If we keep the old numbering, then I am in favor of merging and to generate a single binary, there is no point in pretending we have two. A breaking change in either component is a breaking change in the project. This new scheema automatically reflects this.

I don't know if you saw the file `scripts/release.sh`. There is a `-n` parameter that will tell you what is the next release that it would push. Without this parameter, it pushes the right tag that will trigger the next release. And is protected against "no number change".

Vollstrecker:
> If we keep the old numbering, then I am in favor of merging and to generate a single binary, there is no point in pretending we have two. A breaking change in either component is a breaking change in the project. This new scheema automatically reflects this.

That would also be my favorite, too.

> I don't know if you saw the file `scripts/release.sh`. There is a `-n` parameter that will tell you what is the next release that it would push. Without this parameter, it pushes the right tag that will trigger the next release. And is protected against "no number change".

Nope, releases are your job and I honestly don't care how you design your helpers.

mrjimenez:
> > If we keep the old numbering, then I am in favor of merging and to generate a single binary, there is no point in pretending we have two. A breaking change in either component is a breaking change in the project. This new scheema automatically reflects this.
> 
> That would also be my favorite, too.

Well, my favorite is this PR. To undo and implement the above, there is some work to be done.

> > I don't know if you saw the file `scripts/release.sh`. There is a `-n` parameter that will tell you what is the next release that it would push. Without this parameter, it pushes the right tag that will trigger the next release. And is protected against "no number change".
> 
> Nope, releases are your job and I honestly don't care how you design your helpers.

But that is the whole point, not to screw the release and keep maintainance to the mininum necessary.

PR #610, opened 2026-06-23
refactor: apply misc-include-cleaner across all source files

mrjimenez:
## Summary

- Run `clang-tidy -checks='-*,misc-include-cleaner' -fix` to make direct symbol usage explicit across 75 source files (ixml, upnp, samples, gtest, test programs)
- Improve `scripts/build.sh` with `-a`/`-d`/`-h` flags and align it with CI

## Include cleaner changes

The tool output was reviewed and corrected before committing:

- **Non-portable glibc-internal headers replaced**: `<asm-generic/errno*.h>` → `<errno.h>`, `<asm-generic/socket.h>` / `<bits/sockaddr.h>` → `<sys/socket.h>`, `<bits/types/struct_timeval.h>` → `<sys/time.h>` (inside `#ifndef _WIN32`), `<bits/pthreadtypes.h>` → `<pthread.h>`, `<bits/types/sigset_t.h>` → `<signal.h>`, `<bits/types/struct_sched_param.h>` → `<sched.h>`, `<sys/poll.h>` → removed (already present as `<poll.h>` inside platform guards)
- **`httpparser.c`**: tool placed new includes inside the `#ifndef _GNU_SOURCE` guard — moved them to file scope
- **`upnpapi.c`**: tool removed `<sys/ioctl.h>` from the non-Windows branch — restored (needed for `ioctl()`)
- **`client_table.c`**: tool removed `<stdlib.h>` from inside `#ifdef INCLUDE_CLIENT_APIS` without adding it at file scope — restored at file scope
- **IWYU `pragma: keep`** added to includes that are used via transitive paths clangd cannot trace

## build.sh changes

- Add `-a` flag: enables ASan/LSan and disables integration tests that fail when the static lib carries ASan instrumentation (matches CI `build_asan` job)
- Add `-d` flag: builds source distribution tarball via `cpack -G TBZ2` after tests
- Add `-h` flag: usage
- Default mode now matches CI linux Debug job (`-Wall -Wpedantic -O1 -D_FORTIFY_SOURCE=2`, all integration tests enabled)
- Fix `ctest` invocation: `--test-dir build`, timeout 5→30, add `--exclude-regex '^pthreads4w'`
- Add `post-test.cmake` step; use `-j$(nproc)`; add `set -e`

## Test plan

- [x] CI Build passes on Linux (Debug + Release)
- [ ] CI Build passes on macOS and Windows (portability of replaced headers)
- [x] CI sanitizer build passes (ASan/TSan)
- [x] CI OmniOS build passes (portability of replaced headers)


mrjimenez:
## Bug found and fixed (0a74e867)

The `misc-include-cleaner` refactor introduced a silent type aliasing bug that caused runtime deadlocks in CI (`test-upnp-gena-subscribe-dos` and `test-upnp-ssdp-multicast-recv` timing out).

### Root cause

`ithread_rwlock_t` in `ithread.h` is conditionally defined as either `pthread_rwlock_t` or `pthread_mutex_t`, controlled by `UPNP_USE_RWLOCK` from `autoconfig.h` (pulled in by `config.h`):

```c
#if UPNP_USE_RWLOCK
typedef pthread_rwlock_t ithread_rwlock_t;
#else
typedef ithread_mutex_t ithread_rwlock_t;   /* mutex fallback */
#endif
```

If any translation unit processes `ithread.h` **before** `config.h`, the macro is undefined (evaluates to 0) and `ithread_rwlock_t` silently becomes `pthread_mutex_t` for that entire TU.

`include-cleaner` reordered includes alphabetically in `gena_device.c` and `gena_ctrlpt.c`, placing `ThreadPool.h` (which transitively includes `ithread.h`) before `config.h`:

```c
#include "Callback.h"      // sorts before 'c'
#include "ThreadPool.h"    // pulls in ithread.h HERE — UPNP_USE_RWLOCK not yet defined
...
#include "config.h"        // too late
```

With `UPNP_USE_RWLOCK=0` in those TUs, the static inline `HandleLock()` / `HandleReadLock()` in `upnpapi.h` called `pthread_mutex_lock()` on `GlobalHndRWLock`, which `upnpapi.c` had correctly initialised as a `pthread_rwlock_t`. Mixing the two is undefined behaviour; on Linux it caused an immediate deadlock.

Confirmed via `nm`: the buggy object file referenced `pthread_mutex_lock/unlock`; the original referenced `pthread_rwlock_wrlock/unlock`.

### Fixes

1. **`ithread.h`**: self-includes `config.h` if not already included, making the correct type available regardless of include order and guarding against future regressions.
2. **`gena_device.c`, `gena_ctrlpt.c`**: `config.h` moved to the first `#include`, with an IWYU pragma to prevent include-cleaner from moving it again.
3. **`ThreadPool.c`, `client_table.c`, `sample_util.c`, `test_threadpool_overflow.c`**: same policy fix applied.
4. **`upnp/test/test_ithread_rwlock.c`**: new runtime test — holds a read lock on the main thread and spawns two reader threads; times out (exit 1) if only mutex semantics are available, passes (exit 0) with a real rwlock. Skips (exit 77) on platforms where `UPNP_USE_RWLOCK=0`.

Issue #347, opened 2021-11-21
Normalize HTTP response heades


ngosang:
Gerbera project sent me here: https://github.com/gerbera/gerbera/issues/2229

If I make a request to the Web UI using cURL or Firefox, some of the headers in the response are in upper-case.
That is right according to the HTTP RFC because the headers are case-intensive. This issue has low priority.
Said that, almost all HTTP servers return the headers with capital-case. You have specs here => https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html

```
 curl -vvv http://192.168.1.191:49494/
*   Trying 192.168.1.191:49494...
* Connected to 192.168.1.191 (192.168.1.191) port 49494 (#0)
> GET / HTTP/1.1
> Host: 192.168.1.191:49494
> User-Agent: curl/7.80.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< CONTENT-LENGTH: 25834
< Accept-Ranges: bytes
< CONTENT-TYPE: text/html
< DATE: Sun, 21 Nov 2021 17:47:14 GMT
< LAST-MODIFIED: Sat, 13 Nov 2021 13:07:40 GMT
< SERVER: Linux/5.10.63-v7l+, UPnP/1.0, Portable SDK for UPnP devices/17.7.0
< X-User-Agent: redsonic
< CONNECTION: close
```

mrjimenez:
Thanks for the well-researched report. You're right that RFC 7230 (and the older 2616) defines header names as case-insensitive, but the conventional capitalization — `Content-Length`, `Content-Type`, `Date`, etc. — is what every mainstream HTTP server uses, and deviating from it can confuse naive parsers and tooling.

The uppercase header names are hardcoded in `http_MakeMessage()` (`httpreadwrite.c`) and a few spots in `webserver.c` and `soap_common.c`. The fix is a straightforward string replacement in those three files.

We'll address this in a follow-up commit.

mrjimenez:
Draft PR opened: https://github.com/pupnp/pupnp/pull/559

Branch `chore/issue-347-normalize-http-response-headers` has been created with POC regression tests already committed. The tests are expected to fail until the fix is implemented.

PR #611, opened 2026-07-12
fix: normalize remaining emitted header names to canonical casing

damien78:
## Summary

Normalizes the remaining HTTP header names that the library **emits** in
ALL-CAPS to their canonical casing. This is a follow-up to #347
(commit `4e2c2b7e`), which normalized *response* header names; this PR
covers the **request** and **GENA eventing** paths that #347 didn't touch.

Header names are case-insensitive on the wire, so this is spec-compliant
either way — but some naive peers perform **case-sensitive** string
matching, and the canonical spelling is the interoperable choice.

## Background

RFC 9110 §5.1 (and RFC 7230 §3.2) state that HTTP field names are
case-insensitive, so both `HOST:` and `Host:` are valid. In practice,
however, the near-universal convention is the canonical mixed-case
spelling, and lightweight/embedded HTTP parsers sometimes match field
names literally. #347 already moved the response path to title-case for
exactly this reason; this PR finishes the job for the headers pupnp sends
as a client / event source.

## What changed

Standard HTTP headers → title-case
SOAP header → SOAP-spec casing

## What was deliberately left UPPERCASE (and why)

The UPnP-specific headers are defined in **uppercase** by the UPnP Device
Architecture spec (§1 Discovery/SSDP, §4 Eventing/GENA), and naive UPnP
control points/devices match them case-sensitively against that spelling.
Changing them would risk breaking the exact class of peers this PR is
meant to help, so they are untouched:

- **SSDP** (`ssdp_device.c`, `ssdp_ctrlpt.c`): `HOST`, `NT`, `NTS`, `USN`,
  `ST`, `MX`, `MAN`, `EXT`, `OPT`, `LOCATION`, `CACHE-CONTROL` — these
  travel in HTTPU/HTTPMU datagrams, not HTTP-over-TCP.
- **GENA/SOAP protocol headers**: `NT`, `NTS`, `SID`, `SEQ`, `CALLBACK`,
  `TIMEOUT`, `MAN`, `EXT`.

`upnpapi.c` also keeps a comment reading `CONTENT-TYPE:` — it documents a
Linksys router's non-conforming output (the reason case-insensitive
*parsing* exists), not pupnp's own emission.

## Standards references

| Header | Reference |
|--------|-----------|
| Field names are case-insensitive | RFC 9110 §5.1; RFC 7230 §3.2 |
| `Host` | RFC 9110 §7.2; RFC 7230 §5.4 |
| `Content-Type` | RFC 9110 §8.3; RFC 7231 §3.1.1.5 |
| `Content-Length` | RFC 9110 §8.6; RFC 7230 §3.3.2 |
| `SOAPAction` | SOAP 1.1 (W3C Note, 8 May 2000) §6.1.1 |
| SSDP / GENA header names (left uppercase) | UPnP Device Architecture §1 (Discovery), §4 (Eventing) |

### DIFF
--- a/upnp/src/gena/gena_device.c
+++ b/upnp/src/gena/gena_device.c
@@ -449,8 +449,8 @@ static char *AllocGenaHeaders(
 	const DOMString propertySet)
 {
 	static const char *HEADER_LINE_1 =
-		"CONTENT-TYPE: text/xml; charset=\"utf-8\"\r\n";
-	static const char *HEADER_LINE_2A = "CONTENT-LENGTH: ";
+		"Content-Type: text/xml; charset=\"utf-8\"\r\n";
+	static const char *HEADER_LINE_2A = "Content-Length: ";
 	static const char *HEADER_LINE_2B = "\r\n";
 	static const char *HEADER_LINE_3 = "NT: upnp:event\r\n";
 	static const char *HEADER_LINE_4 = "NTS: upnp:propchange\r\n";
--- a/upnp/src/genlib/net/http/httpreadwrite.c
+++ b/upnp/src/genlib/net/http/httpreadwrite.c
@@ -903,7 +903,7 @@ int http_Download(const char *url_str,
 		HTTPMETHOD_GET,
 		url.pathquery.buff,
 		url.pathquery.size,
-		"HOST: ",
+		"Host: ",
 		hoststr,
 		hostlen);
 	if (ret_code != 0) {
@@ -1069,7 +1069,7 @@ int MakeGenericMessage(http_method_t method,
 				1,
 				"s"
 				"bcDCU",
-				"HOST: ",
+				"Host: ",
 				hoststr,
 				hostlen);
 		}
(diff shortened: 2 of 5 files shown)


Vollstrecker:
Just to get that right: The changes are to help parsers that are not spec-conform? For #347 I was a bit bit slow, but imho it's better to let these poorly designed parsers run into errors so they can be made spec-conform instead of supporting their broken design.

damien78:
Such poorly designed parsers I've encountered are inside hardware audio UPnP renderers (called hifi "streamers"). 
So sending canonical mixed-case headers is much easier to do (and makes no harm to any other), than leaving their users stuck trying to get the manufacturer of their poorly designed box release a firmware update.

mrjimenez (approved):
Verified the one spot where a case-sensitive string match could have broken behavior: `add_man_header()` in `soap_ctrlpt.c` does `strstr(headers->buf, "SOAPAction:")` to splice in the `MAN:` header for the M-POST retry path — the header is built earlier with the same new casing (`"SOAPAction: \""`), so the match still succeeds. Confirmed the SSDP/GENA protocol headers you intentionally left uppercase are indeed untouched, matching the PR description's rationale. Full CI build matrix (Linux/macOS/Windows/MSys/OmniOS), sanitizers, TSan, and lint all pass. LGTM.

mrjimenez:
Thanks for the thorough write-up and standards references, @damien78 — made this easy to verify and merge. Nice follow-up to #347, and appreciated that you called out exactly which headers were deliberately left uppercase and why.

PR #612, opened 2026-07-12
feat: allow Windows interface selection by adapter name (GUID)

damien78:
## Summary

Adds an opt-in build option, `UPNP_ADAPTER_UUID_NAME` (default `OFF`), that
changes how `UpnpGetIfInfo()` identifies a network interface on Windows: by the
adapter's **AdapterName** (its GUID) instead of its **FriendlyName**.

## Motivation

On Windows, `UpnpGetIfInfo()` matches the requested interface against the
adapter's `FriendlyName` (e.g. "Ethernet", "Wi‑Fi", or a localized/renamed
label). Two problems follow from this:

1. **Lossy conversion.** `FriendlyName` is a wide-character (UTF‑16) string,
   but `gIF_NAME` is a narrow `char` buffer, so the code converts it with
   `wcstombs_s`. For adapters whose friendly name contains non-ASCII or
   localized characters, the conversion can be incomplete or fail, and the
   adapter then never matches — the interface effectively can't be selected.
   The existing code even documents this as a "partial fix … which will work
   with many (but not all) adapters."

2. **Unstable identifier.** Friendly names are user-editable and locale
   dependent. They can change when the user renames an adapter or switches the
   system language, which makes them a poor key for deterministic,
   reproducible interface selection.

`AdapterName`, by contrast, is a stable ASCII GUID string
(`{XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX}`) provided as `PCHAR`. It needs no
wide→narrow conversion, is unique, and does not change across renames or
locale changes. Applications that already obtain the adapter GUID from the
Windows networking APIs can then pass it straight to `UpnpGetIfInfo()` and get
deterministic selection.

## Behavior & compatibility

- Default is `OFF`; existing builds and the current FriendlyName behavior are
  completely unchanged.
- When enabled, callers pass the adapter GUID as the interface name.
- Affects Windows only; on other platforms the option has no effect.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -171,6 +171,12 @@ jobs:
       matrix:
         sys: [ mingw64, mingw32, ucrt64, clang64 ]
         conf: [ Release, Debug ]
+        include:
+          # Exercise the opt-in UPNP_ADAPTER_UUID_NAME option on one
+          # representative toolchain/config combo.
+          - sys: mingw64
+            conf: Release
+            extra_cmake_flags: -DUPNP_ADAPTER_UUID_NAME=ON
     defaults:
       run:
         shell: msys2 {0}
@@ -209,6 +215,7 @@ jobs:
             -DCMAKE_BUILD_TYPE=${{ matrix.conf }} \
             -DIXML_ENABLE_TESTING_INTEGRATION=OFF \
             -DUPNP_ENABLE_TESTING_INTEGRATION=OFF \
+            ${{ matrix.extra_cmake_flags }} \
             -DCMAKE_INSTALL_PREFIX=${{ GITHUB.WORKSPACE }}/test-install
       - name: Build
         run: cmake --build build --config ${{ matrix.conf }}
--- a/cmake/autoconfig.h.in
+++ b/cmake/autoconfig.h.in
@@ -49,6 +49,10 @@
 /* see upnpconfig.h */
 #cmakedefine UPNP_MINISERVER_REUSEADDR 1
 
+/* On Windows, select the network interface by its adapter name (GUID) instead
+ * of the friendly name (see UpnpGetIfInfo in upnp/src/api/upnpapi.c). */
+#cmakedefine UPNP_ADAPTER_UUID_NAME 1
+
 /* Type for storing the length of struct sockaddr */
 #cmakedefine socklen_t 1
 
--- a/cmake/options.cmake
+++ b/cmake/options.cmake
@@ -9,6 +9,7 @@ option (IXML_ENABLE_SCRIPT_SUPPORT "script support for IXML document tree, see i
 option (IXML_ENABLE_TESTING "enable tests for ixml" ${testing_default})
 option (IXML_ENABLE_TESTING_INTEGRATION "enable integrationtests for ixml" ${testing_default})
 
+option (UPNP_ADAPTER_UUID_NAME "on Windows, select the network interface by its adapter name (GUID) instead of the friendly name" OFF)
 option (UPNP_BUILD_SAMPLES "compilation of upnp/sample/ code" ON)
 option (UPNP_BUILD_SHARED "Build shared libraries" ON)
 option (UPNP_BUILD_STATIC "Build static libraries" ON)
--- a/upnp/inc/upnpconfig.h.cm
+++ b/upnp/inc/upnpconfig.h.cm
@@ -129,4 +129,9 @@
  * with SO_REUSEADDR (i.e. configure --enable_reuseaddr) */
 #cmakedefine UPNP_MINISERVER_REUSEADDR 1
 
+/** Defined to 1 if, on Windows, the library has been compiled to select the
+ *  network interface by its adapter name (GUID) instead of the friendly name
+ *  (i.e. cmake -DUPNP_ADAPTER_UUID_NAME=ON) */
+#cmakedefine UPNP_ADAPTER_UUID_NAME 1
+
 #endif /* UPNP_CONFIG_H */
--- a/upnp/src/api/upnpapi.c
+++ b/upnp/src/api/upnpapi.c
@@ -3826,6 +3826,12 @@ int UpnpGetIfInfo(const char *IfName)
 		}
 		if (ifname_found == 0) {
 			/* We have found a valid interface name. Keep it. */
+	#ifdef UPNP_ADAPTER_UUID_NAME
+			strncpy(gIF_NAME,
+				adapts_item->AdapterName,
+				sizeof(gIF_NAME) - 1);
+			gIF_NAME[sizeof(gIF_NAME) - 1] = 0;
+	#else
 			/*
 			 * Partial fix for Windows: Friendly name is wchar
 			 * string, but currently gIF_NAME is char string. For
@@ -3840,9 +3846,18 @@ int UpnpGetIfInfo(const char *IfName)
 				adapts_item->FriendlyName,
 				sizeof(gIF_NAME));
 			free(s);
-
+	#endif
 			ifname_found = 1;
 		} else {
+	#ifdef UPNP_ADAPTER_UUID_NAME
+			if (strncmp(gIF_NAME,
+				    adapts_item->AdapterName,
+				    sizeof(gIF_NAME)) != 0) {
+				/* This is not the interface we're looking for.
+				 */
+				continue;
+			}
+	#else
 			/*
 			 * Partial fix for Windows: Friendly name is wchar
 			 * string, but currently gIF_NAME is char string. For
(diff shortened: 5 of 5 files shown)


Vollstrecker:
As far as I understand the code, I see no potential to break existing builds, looks more like if present use it. I don't see any reason to make it optional.

mrjimenez (dismissed):
Thanks for this — matching on `AdapterName` instead of the wchar `FriendlyName` is a real improvement, and the `strncpy`/`strncmp` bounds handling here is solid. Three things need to be resolved before merge; I don't have "allow edits from maintainers" enabled on this branch so I can't push directly — patch attached below (`git apply` should take it cleanly), or enable maintainer edits on the PR and I'll push a fixup + CI-matrix commit myself.

### 1. clang-format violation (blocking — CI's `lint-clang-format` job will fail)
This repo's `.clang-format` uses `IndentPPDirectives: BeforeHash`, so the new `#ifdef UPNP_ADAPTER_UUID_NAME` / `#else` / `#endif` need a tab before the `#`, and the new `strncmp(...)` argument list needs deeper continuation alignment. Verified against `clang-format-19` (CI's pinned version) with `main` confirmed clean on this file — this is new, not pre-existing.

### 2. `UPNP_ADAPTER_UUID_NAME` isn't exposed in the installed public header
Every other `UPNP_HAVE_*`/`UPNP_ENABLE_*` build option is mirrored in `upnp/inc/upnpconfig.h.cm` (which CMake installs as `upnpconfig.h`) so applications linking against the library can detect at compile time which optional features it was built with. Since this option changes what format callers must pass to `UpnpGetIfInfo()`, it needs the same treatment — otherwise consumers have no portable way to know which mode is in effect.

### 3. No CI coverage for `UPNP_ADAPTER_UUID_NAME=ON`
The whole Windows branch of `UpnpGetIfInfo()` is `#ifdef _WIN32`-gated, so it's never compiled on Linux CI, and none of the Windows jobs (`windows-msvc`, `windows-msys2`) currently build with the flag on — so a latent compile error in the new branch would ship undetected. Added it to exactly one `windows-msys2` matrix combo (`mingw64` + `Release`) via a matrix `include:` override, so the other 7 combos keep testing the shipped default (`OFF`) untouched.

---

```diff
diff --git a/.github/workflows/ccpp.yml b/.github/workflows/ccpp.yml
index 96aee682..66656157 100644
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -171,6 +171,12 @@ jobs:
       matrix:
         sys: [ mingw64, mingw32, ucrt64, clang64 ]
         conf: [ Release, Debug ]
+        include:
+          # Exercise the opt-in UPNP_ADAPTER_UUID_NAME option on one
+          # representative toolchain/config combo.
+          - sys: mingw64
+            conf: Release
+            extra_cmake_flags: -DUPNP_ADAPTER_UUID_NAME=ON
     defaults:
       run:
         shell: msys2 {0}
@@ -209,6 +215,7 @@ jobs:
             -DCMAKE_BUILD_TYPE=${{ matrix.conf }} \
             -DIXML_ENABLE_TESTING_INTEGRATION=OFF \
             -DUPNP_ENABLE_TESTING_INTEGRATION=OFF \
+            ${{ matrix.extra_cmake_flags }} \
             -DCMAKE_INSTALL_PREFIX=${{ GITHUB.WORKSPACE }}/test-install
       - name: Build
         run: cmake --build build --config ${{ matrix.conf }}
diff --git a/upnp/inc/upnpconfig.h.cm b/upnp/inc/upnpconfig.h.cm
index b4c47480..cf043c13 100644
--- a/upnp/inc/upnpconfig.h.cm
+++ b/upnp/inc/upnpconfig.h.cm
@@ -129,4 +129,9 @@
  * with SO_REUSEADDR (i.e. configure --enable_reuseaddr) */
 #cmakedefine UPNP_MINISERVER_REUSEADDR 1
 
+/** Defined to 1 if, on Windows, the library has been compiled to select the
+ *  network interface by its adapter name (GUID) instead of the friendly name
+ *  (i.e. cmake -DUPNP_ADAPTER_UUID_NAME=ON) */
+#cmakedefine UPNP_ADAPTER_UUID_NAME 1
+
 #endif /* UPNP_CONFIG_H */
diff --git a/upnp/src/api/upnpapi.c b/upnp/src/api/upnpapi.c
index 5708d089..185d2c66 100644
--- a/upnp/src/api/upnpapi.c
+++ b/upnp/src/api/upnpapi.c
@@ -3826,12 +3826,12 @@ int UpnpGetIfInfo(const char *IfName)
 		}
 		if (ifname_found == 0) {
 			/* We have found a valid interface name. Keep it. */
-#ifdef UPNP_ADAPTER_UUID_NAME
+	#ifdef UPNP_ADAPTER_UUID_NAME
 			strncpy(gIF_NAME,
 				adapts_item->AdapterName,
 				sizeof(gIF_NAME) - 1);
 			gIF_NAME[sizeof(gIF_NAME) - 1] = 0;
-#else
+	#else
 			/*
 			 * Partial fix for Windows: Friendly name is wchar
 			 * string, but currently gIF_NAME is char string. For
@@ -3846,18 +3846,18 @@ int UpnpGetIfInfo(const char *IfName)
 				adapts_item->FriendlyName,
 				sizeof(gIF_NAME));
 			free(s);
-#endif
+	#endif
 			ifname_found = 1;
 		} else {
-#ifdef UPNP_ADAPTER_UUID_NAME
+	#ifdef UPNP_ADAPTER_UUID_NAME
 			if (strncmp(gIF_NAME,
-				adapts_item->AdapterName,
-				sizeof(gIF_NAME)) != 0) {
+				    adapts_item->AdapterName,
+				    sizeof(gIF_NAME)) != 0) {
 				/* This is not the interface we're looking for.
 				 */
 				continue;
 			}
-#else
+	#else
 			/*
 			 * Partial fix for Windows: Friendly name is wchar
 			 * string, but currently gIF_NAME is char string. For
@@ -3881,7 +3881,7 @@ int UpnpGetIfInfo(const char *IfName)
 				 */
 				continue;
 			}
-#endif
+	#endif
 		}
 		/* Loop thru this adapter's unicast IP addresses. */
 		uni_addr = adapts_item->FirstUnicastAddress;
```

damien78:
thanks for the thorough review. I've enabled edits by maintainer on the PR so you can push the fixes

mrjimenez:
@Vollstrecker good question — I think it's worth keeping this opt-in, and I'd like to explain why, since "won't break existing builds" and "won't break existing behavior" are two different guarantees here.

The change touches two distinct code paths in `UpnpGetIfInfo()`:

1. **Auto-select, no `IfName` given** (`ifname_found == 0`): here, unconditionally switching to `AdapterName` would indeed be harmless — it only changes what gets copied into `gIF_NAME` for internal bookkeeping/reporting, nothing external depends on that value's *format*.

2. **Explicit match, caller passes an `IfName`** (`ifname_found == 1`, the `else` branch): this is where an unconditional switch would be a real behavioral break. Today, any existing application on Windows that calls `UpnpGetIfInfo("Ethernet")` (or whatever friendly name it obtained through its own UI/config) works, because the code matches against `FriendlyName`. If we unconditionally switch the comparison to `AdapterName`, that same call would silently match *no* adapter, because it's now comparing the caller's friendly-name string against a GUID. Nothing fails to compile or link — it just stops finding interfaces at runtime, for every existing caller that passes a name. That's a public API contract change (what format `IfName` must be in), not an additive feature.

This library already has precedent for exactly this situation: options like `UPNP_ENABLE_BLOCKING_TCP_CONNECTIONS`, `UPNP_MINISERVER_REUSEADDR`, and `UPNP_ENABLE_UNSPECIFIED_SERVER` all default to `OFF` specifically because they alter existing behavior rather than just adding something new — the convention here is "don't silently change behavior for callers who haven't opted in." I think `UPNP_ADAPTER_UUID_NAME` fits that same pattern, so I'd keep it opt-in/default-`OFF` as originally submitted.

mrjimenez (approved):
All three items resolved in the latest commit:

- `lint-clang-format`: now passing
- `UPNP_ADAPTER_UUID_NAME` now mirrored in `upnp/inc/upnpconfig.h.cm`
- `windows-msys2` matrix now exercises `-DUPNP_ADAPTER_UUID_NAME=ON` on `mingw64`/`Release`, and that job passed, confirming the new code path actually compiles on a real Windows toolchain

Full CI (Build, CodeQL, lint-clang-format) is green. Approving.

Vollstrecker:
The entry in UPNP.cmake.in is also missing if it is kept as an option.

mrjimenez:
@Vollstrecker you're right, good catch — `UPNP.cmake.in` has the same pattern as `upnpconfig.h.cm` (every build option mirrored as a CACHE variable for `find_package(UPNP)` consumers) and I missed it there. Opened #615 to fix it.

PR #613, opened 2026-07-12
ixml merge Proposal 1

Vollstrecker:
As we talked about the recent changes, merging of ixml into upnp came on the tablet.

This could be one way to do this.

With this ixml get's the same version as upnp and migrates to be a component of it. find_package(IXML) no longer works, it's found as part of find_package(UPNP) or if only that is needed COMPONENTS IXMLShared or IXMLStatic can be used.

Target aliases have been changed from IXML::shared to UPNP::IMXLShared and the same for static. I didn't implement any compat-layer or warnings as you mentioned it's only used by libupnp. I guess we'll find out if someone complains.

I thought about UPNP::IXML::Shared for a short period, but for staying coherent this would also mean to change UPNP::Shared to UPNP::UPNP::Shared. Even better PUPNP::UPNP::Shared and PUPNP::IXML::Shared as the package should really be called as the project. That would also mean that everything using it now would break, and also that it's even more coherent to switch the package_name so it would have to be found via find_package(PUPNP), which also would make compat-stuff on users side needed.

I didn't adjust the docs or all the scripts atm. as this is just to provide something to talk about.


mrjimenez:
AI generated review:

**CI: OmniOS Release failure isn't flaky — it's a deterministic regression from this diff, and restarting won't fix it.**

All four failures (`ixml_add_subdirectory_build_{shared,static}`, `upnp_add_subdirectory_build_{shared,static}`) time out at exactly ~30.0–30.05s, every run. Debug passes but already spends 27–28s of the same 30s budget on these same tests — this was already on a knife's edge before Release tipped it over.

**Root cause:** each `_configure_{shared,static}` test passes exactly one flavor explicitly, e.g. for `upnp_add_subdirectory_configure_shared`:
```
-DUPNP_BUILD_SHARED=ON
```
`UPNP_BUILD_STATIC` is never passed via `-D` here — it was always left to the template's own `option()` default. Before this PR that default was hardcoded `OFF`, so only the shared flavor got compiled inside the nested `add_subdirectory()` build. This PR changes the template to:
```diff
-option(UPNP_BUILD_STATIC "" OFF)
+option(UPNP_BUILD_STATIC "" @UPNP_BUILD_STATIC@)
```
`@UPNP_BUILD_STATIC@` gets substituted with the *outer* build's actual value, which is `ON` by default project-wide (`cmake/options.cmake`) and never overridden on OmniOS CI. So the "shared" test now silently also builds the static flavor (and vice versa for the "static" test) — roughly doubling the compile work in each of these four tests, for a scenario nobody is actually asking to test.

**Suggested fix** — make the counterpart flavor explicit rather than implicit, in the same four spots that already pass one `-D` flag:

`upnp/test/CMakeLists.txt`:
```diff
 			-DBUILD_TESTING=OFF
 			-DUPNP_BUILD_SHARED=ON
+			-DUPNP_BUILD_STATIC=OFF
 			-G "${CMAKE_GENERATOR}"
 			-S${CMAKE_CURRENT_BINARY_DIR}/add_subdirectory_test)
```
(`upnp_add_subdirectory_configure_shared`), and symmetrically `-DUPNP_BUILD_SHARED=OFF` added to `upnp_add_subdirectory_configure_static`.

`ixml/test/CMakeLists.txt`: same two additions to `ixml_add_subdirectory_configure_shared` / `ixml_add_subdirectory_configure_static`.

This keeps your intent (real option values flowing through instead of a hardcoded no-op) while restoring each test to building the one flavor it's named for. Should bring these back under budget without touching the (global, suite-wide) `--timeout 30` flag.

Same root cause exists in #614 (identical template diff, same timeout there too), fyi.

Vollstrecker:
Looks better now. You also updated the part about configure.ac in the cmake-migration doc. I guess that part can go completely.

I also think about renaming IXML_SCRIPT_SUPPORT option, as it's the only option left that influences ixml. Usually cmake-gui groups the options based on the first part, so all UPNP_X get into a group that makes chainbuilds easier to configure as all options of a package are grouped together, the single ixml option would got somewhere else.

And maybe would should update the readme also, as it still mentions BUILD_TESTING which isn't used anymore, this functionality was split into UPNP_BUILD_TESTING and UPNP_BUILD_TESTING_INTEGRATION.

Vollstrecker:
Found it, it get's mapped to HAVE_SCRIPT_SUPPORT which wasn't migrated to the config.

Vollstrecker:
I hate conflicts. Should be good to go now.

mrjimenez:
Heads up: force-pushed `proposal-1` just now — flattened the merge commit ("Merge branch 'main' into proposal-1") back into a linear history, so `gh`/GitHub's rebase-and-merge option stays available when this is ready to land (it gets disabled once a branch contains a merge commit).

No content changed — verified the resulting tree is byte-for-byte identical to what you had before flattening it (diffed the two, zero lines). Also picked up two `main` commits that had landed after the snapshot your merge grabbed, so you get those for free too. Full build + test suite green locally before pushing.

If you'd already pulled the previous tip locally, you'll want `git fetch && git reset --hard origin/proposal-1` (or just re-clone the branch) rather than merging/rebasing your own copy against it — the history's genuinely different now, not just diverged.

Sorry for rewriting something you just finished untangling — figured better to flag it clearly than have it surprise you on the next pull.

mrjimenez:
Looks fine! And CI is green.

Are you ok with that? You did noticed that according to this PR, the new pupnp version is 22.0.3, right?

Vollstrecker:
As I get noticed on changes, I'm aware of the changes. And as I have ~30 old branches I guess a new clone is overdue.

According to the CMakeLists.txt L15-17 from main, that's the right version.

PR #614, opened 2026-07-12
ixml merge Proposal 2

Vollstrecker:
As proposal one but more radical. This one removes ixml completely as lib on it's own and integrates it into libupnp.

While writing these two, I saw autoconfig.h.cm and autoconfig.h.in, the .cm is used, I guess the in is from the autoconf era.


Vollstrecker:
I guess for this we need to think about proper export of the ixml symbols.

PR #620, opened 2026-07-13
Follow-up cleanup for proposal-1 (docs, README, option naming)

mrjimenez:
## Summary

Addresses the three follow-up points from your reply on #613:

- **cmake-migration.md**: dropped the leftover "Previously the authoritative version was in `configure.ac`..." lead-in from the version-numbers section. That framing belonged to the original autotools migration; this section is about the umbrella-math → single-version change and doesn't need it.
- **README.md**: `BUILD_TESTING` was still documented as a single option; replaced with the two that actually exist now, `UPNP_ENABLE_TESTING` and `UPNP_ENABLE_TESTING_INTEGRATION`.
- **cmake/options.cmake**: renamed `IXML_ENABLE_SCRIPT_SUPPORT` → `UPNP_IXML_ENABLE_SCRIPT_SUPPORT` — it was the last `IXML_`-prefixed option, so it was sorting away from the rest in cmake-gui instead of grouping with the other `UPNP_*` options.

One thing deliberately left alone: `IXML_HAVE_SCRIPTSUPPORT`, the macro this option feeds. It's part of the public `ixml.h` contract — consumers `#ifdef` against it directly in their own code — so it's orthogonal to the build-option naming cleanup and renaming it would be an unrelated breaking change for downstream users. Only the option name moved.

Also noticed while in `cmake/options.cmake`: the `ENABLE_TESTING` deprecation-migration message (lines 3-4) still mentions `IXML_ENABLE_TESTING`, which no longer exists as an option since the merge. Left as-is since it's outside what was asked here — happy to clean it up separately if wanted.

## Test plan

- [x] Full reconfigure + build + `ctest` (116 tests) — all pass, no regressions
- [x] Confirmed `IXML_HAVE_SCRIPTSUPPORT` still ends up in `compile_commands.json` after the option rename (macro plumbing intact)
- [x] Re-ran the four `add_subdirectory_build_{shared,static}` tests — still 9-14s each


Vollstrecker:
Two things to that one:

1. I can't find it, but I'm sure IXML_SCRIPT_SUPPORT does something, somewhere, or is it only for setting defines?
2. I'm pretty sure you could commit directly to the branch.

mrjimenez:
Good catch to double-check rather than just take the rename at face value — turns out it's very much not "just a define"! `IXML_HAVE_SCRIPTSUPPORT` gates real stuff in the public `ixml.h`:

- A field in the public `IXML_Node` struct itself: `void *ctag; /* custom tag */` — so it actually changes the struct's layout/size, not just which functions are visible.
- Two exported functions: `ixmlNode_setCTag()` and `ixmlSetBeforeFree()`.
- A typedef, `IXML_BeforeFreeNode_t`.

So it's a genuine ABI-affecting feature switch (struct layout changes depending on the setting!), not dead weight — good thing we didn't just leave it out.

Side note, not something to act on here, just flagging for later: since it changes struct layout, a `libixml` built with it ON isn't ABI-compatible with one built OFF, and the current SOVERSION scheme doesn't distinguish the two. Pre-existing situation though, nothing this PR introduces.

And yep, noted on committing directly — will do that from here on out instead of routing everything through PRs into your branch. Thanks for the flexibility on that, makes this a lot smoother.

PR #629, opened 2026-09-02
OSS-Fuzz: Add new fuzzer targets service table processing

arthurscchan:
This PR creates a new OSS-Fuzz fuzzer targets service table processing, together with the build fix in OSS-Fuzz (https://github.com/google/oss-fuzz/pull/16102) to enable the new fuzzer.

### DIFF
--- a/fuzzer/CMakeLists.txt
+++ b/fuzzer/CMakeLists.txt
@@ -3,6 +3,7 @@ cmake_minimum_required(VERSION 3.11)
 project(fuzzer LANGUAGES C)
 add_executable(FuzzIxml FuzzIxml.c)
 add_executable(FuzzHttp FuzzHttp.c)
+add_executable(FuzzServiceTable FuzzServiceTable.c)
 
 if(NOT DEFINED LOCAL_RUN)
 	message("LOCAL_RUN is not defined")
@@ -11,6 +12,8 @@ if(NOT DEFINED LOCAL_RUN)
 	target_link_libraries(FuzzIxml ixml_static ${LIB_FUZZING_ENGINE})
 	target_include_directories(FuzzHttp PRIVATE ${CMAKE_SOURCE_DIR}/upnp/src/threadutil)
 	target_link_libraries(FuzzHttp upnp_static ${LIB_FUZZING_ENGINE})
+	target_include_directories(FuzzServiceTable PRIVATE ${CMAKE_SOURCE_DIR}/upnp/src/threadutil)
+	target_link_libraries(FuzzServiceTable upnp_static ${LIB_FUZZING_ENGINE})
 else()
 	message("LOCAL_RUN is ${LOCAL_RUN}")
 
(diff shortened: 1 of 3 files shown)


mrjimenez:
Thanks Arthur — this is a good target to add. `getServiceTable()` sits right at the end of the control-point path that consumes a remote device description, so fuzzing it is worthwhile, and the CMake wiring follows `FuzzHttp` exactly.

I reviewed and verified it locally with clang-19 (`-DFUZZER=ON`, ASan + libFuzzer):

- builds clean, no warnings;
- 60 s run with an XML seed: ~1.5 M executions at ~24k/s, no crashes and no leaks.

I found one build problem and pushed a fix to your branch directly (377ffe4b), since maintainer edits are enabled:

`service_table`, `getServiceTable()` and `freeServiceTable()` are declared inside `#ifdef INCLUDE_DEVICE_APIS` / `#if EXCLUDE_GENA == 0` in `upnp/src/inc/service_table.h`. A client-only build therefore fails to compile the target:

```
$ cmake -DFUZZER=ON -DUPNP_ENABLE_DEVICE_API=OFF ...
FuzzServiceTable.c:13:2: error: use of undeclared identifier 'service_table'
FuzzServiceTable.c:31:3: error: call to undeclared function 'getServiceTable'
FuzzServiceTable.c:32:3: error: call to undeclared function 'freeServiceTable'
7 errors generated.
```

OSS-Fuzz builds with the defaults so it would never have hit this, but the fuzzers should stay buildable in every configuration. The body is now compiled only when both features are present. Both configurations build clean afterwards, and the fuzzer behaves the same.

The same commit adds two comments: one noting that `ixml.h` is not included directly because `service_table.h` already pulls it in, and one recording that the `Size > 65536` test is only a harness-side safety net — the real input bound comes from libFuzzer's `-max_len`, which is set neither on the command line nor through a `FuzzServiceTable.options` file, so libFuzzer's 4096-byte default applies and that test never fires in practice. If you want the target to see larger device descriptions, adding

```ini
[libfuzzer]
max_len = 65536
```

to a `FuzzServiceTable.options` file in google/oss-fuzz#16102 would be the place to do it. Not a blocker for this PR.

One last note, unrelated to the fuzzer: the formatting fix to `upnp/test/poc_ghsa_f8r2_nullbyte.c` is correct and welcome — `main` is currently red on the clang-format lint because of exactly those two hunks, so that commit fixes my tree, not yours. Thank you.

Approving and merging.

mrjimenez (approved):
Verified locally: builds clean with clang-19 under libFuzzer + ASan, 60 s run with an XML seed gives ~1.5 M executions with no crashes and no leaks. The client-only build breakage is fixed in 377ffe4b. Good addition — thanks.

arthurscchan:
Hi @mrjimenez, thanks for fixing the minor building error for me and sorry to overlook on general building for the fuzzers. For the larger max_len, I think it is better to let OSS-Fuzz run for a bit before consider a larger max_len. Thanks again for your help and approval of the PR. I do on the way to develop more fuzzers and will possibly have more PRs ready in the coming days. Thanks.

mrjimenez:
Thanks Arthur! No need to apologise, the fuzzers aren't covered by any CI build here, so that configuration was never going to be caught by anything but a manual check.

On the max_len thing, I agree, better to see real coverage data from OSS-Fuzz before deciding whether the target is actually input-size starved.

